PR #25, opened 2020-08-05
Add contribution guidelines

tfidfwastaken:
I have added contribution guidelines that, when followed will ensure
less BT for us and a clean commit history for everyone involved

By the simple rebase operations outlined, it will also help our
contributors feel confident in other real-world contribution scenarios
where the maintainer expects a clean set of commits in the commit
history of the PR

Please do check if this is not too bhaari to understand :p

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,51 @@
+# Contributing Guidelines
+
+- Look for issues in the repository. Identify an issue you would like to work on.
+- Once you have done that, leave a comment on the issue saying that you want to work on it. The maintainers will give you the go-ahead.
+- Fork the repository, make your changes on the `master` branch, or on the appropriate feature branch if applicable
+- Clone the fork on your system. Add a remote to the original `pesos/grofer` repository called `upstream`
+```
+git clone https://github.com/<your-github-username>/grofer.git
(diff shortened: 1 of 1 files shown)


tfidfwastaken:
Oof, just as I pushed I remembered that github has an option for "squash and merge" in the arrow next to "merge pull request".

So now there is a tradeoff. Should I simplify the instruction where I mention squashing, so the PR can have be a littered mess and we deal with it and make their change into one commit using GitHub's "squash and merge"?

The latter will make it easier for contributors, but the former will teach them more, which can be useful when contributing to other projects

MadhavJivrajani:
Since this project is starting to get attention from people who are not a part of PESOS, according to me, ease of contribution should be a priority, square and merge seems like the better option.

tfidfwastaken:
Alright, I will remove the mention of squashes. I think our main priority should be accessibility.

MadhavJivrajani:
Will wait for @Gituser143 and @metonymic-smokey 's opinion before confirming merge

tfidfwastaken:
Okay I have made the instructions a little more simpler. Feel free to squash and merge these two commits when you are ready ;)

MadhavJivrajani:
Yeah, I'll enable that

PR #29, opened 2020-08-06
added user defined UI refresh rate

### LINKED ISSUE #15: [FEATURE REQ] provide functionality for user defined refresh rates
Currently, the UI refreshes at a default, hard-coded rate of 100 ms. A user-defined refresh rate would be beneficial. 

Add a persistent flag to the `root` command in `cmd/root.go` called `-r` or `--refresh` in long-form with a default value of 100 (ms). 

**Additional context**
All commands created using the `cobra` cli tool are by default children of the `root` command, by setting a persistent flag on `root`, it will allow `grofer` to use the same flag for all child commands as well and a similar syntax for refresh rate can be used for all commands defined. 

For more information, refer to this [link](https://github.com/spf13/cobra#working-with-flags).

Ex:
```
grofer proc -r 10
grofer -r 1000
grofer proc -p PID -r 1000
```

yongruilin:
# Description

Fixes #15 

## Type of change

# Checklist:


MadhavJivrajani:
Most of this looks good, I had a few questions:
- Where is the value of the refresh rate being extracted? I see a global variable called refresh rate which is being passed even to the `process` functions. 
- Any reason as to why the `proc` does not have its own flag for refresh rate as the flag set in `root` is non-persistent?

yongruilin:
@MadhavJivrajani Thanks for your comment!
> * Where is the value of the refresh rate being extracted? I see a global variable called refresh rate which is being passed 
even to the `process` functions.
> * Any reason as to why the `proc` command does not have its own flag for refresh rate as the flag set in `root` is non-persistent?

It is extracted in the init() function in cmd/root.go as a persistent flag. Since it is used in both root and proc command, is it better to put it as a persistent flag?
> * Is the `DefaultRefreshRate` const being used anywhere?

Yes, or should the default value be different between the proc and root?

MadhavJivrajani:
- In the `init()` function, the line 
```go
rootCmd.Flags().Int32P(...)
```
only sets the flag to a default value of 1000, but if you wanted to something like:
```
grofer -r 2000
```
where would it read the value of 2000? Or am I missing something? To extract the value of a flag passed in the command line, something similar to [this](https://github.com/pesos/grofer/blob/8a262afb969b7e4d8a6c04df2efd25fdbda00a68/cmd/proc.go#L50) would need to be done.

- In principle, yes for the UI of both commands `grofer` and `grofer proc [-p PID]` the same refresh rate can be used, but its better to set a separate flag in the `proc` command as you did in `root` and extract the value if you aren't doing that and pass that extracted value to the functions. 

- I don't think you need a `DefaultRefreshRate` const to take care of the default refresh rate case. If you look at `cmd/proc.go` the default value for the `-p` flag is -1. So if you were to run 
```
grofer proc
```
and not pass any `-p` flag, the value extracted in `RunE` for that command will default to -1, so you can just pass the extracted value after checking for the refresh rate to be at least 1000 ms.

so if you were to run:
```
grofer proc -p PID
```
and not pass a `-r` flag to it, then the extracted value would default to 1000 ms

Please feel free to ask any questions you may have! :smile:

yongruilin:
@MadhavJivrajani 
> where would it read the value of 2000? Or am I missing something? To extract the value of a flag passed in the command line, something similar to [this](https://github.com/pesos/grofer/blob/8a262afb969b7e4d8a6c04df2efd25fdbda00a68/cmd/proc.go#L50) would need to be done.

Please check the code in the init function of root.go
```
rootCmd.PersistentFlags().Int32VarP(&refreshRate, "refresh", "r", DefaultRefreshRate, "UI refreshes rate in milliseconds greater than 1000")
```
this will pass the value from the command line with flag `--refresh`  or `-r`
Since it is a persistent flag, it is also bound to the children command `proc`

If we don't pass a `-r` flag, it will use the default value (to both root command and children commands) 
[cobra command api](https://github.com/spf13/pflag/blob/81378bbcd8a1005f72b1e8d7579e5dd7b2d612ab/int32.go#L60)

yongruilin:
> * [ ]  Change refresh rate to a non-persistent flag and add a separate refresh rate non-persistent flag to `proc`

To avoid panic we should use the persistent flag. Otherwise, we should introduce another flag name for the children command
> * [ ]  Remove the `DefaultRefreshRate` const

I think any hard-coded value is better to be defined in const to avoid confusion. If we need different default value for different  children commands, we can define their `Run` or `RunE` functions

yongruilin:
> * [x]  Change refresh rate to a non-persistent flag and add a separate refresh rate non-persistent flag to `proc`

Thanks for your suggestion, please check the update.
> * [ ]  Remove the `DefaultRefreshRate` const

I defined different default values for overall and process. In the future, we can also define different default values for other subcommands. Better to have a const for hard-coded value to avoid confusion.

cmd/proc.go
```diff
@@ -20,14 +20,18 @@ import (
...
 	"github.com/pesos/grofer/src/process"
-	proc "github.com/shirou/gopsutil/process"
-	"github.com/spf13/cobra"
+	"github.com/pesos/grofer/src/utils"
 )
 
+const (
+	DefaultProcRefreshRate = 1000
+)
+var procRefreshRate int32
```
MadhavJivrajani, line 34:
```suggestion

```

cmd/proc.go
```diff
@@ -92,5 +97,6 @@ func init() {
 	// Cobra supports local flags which will only run when this command
 	// is called directly, e.g.:
 	// procCmd.Flags().BoolP("toggle", "t", false, "Help message for toggle")
+	procCmd.Flags().Int32VarP(&procRefreshRate, "refresh", "r", DefaultProcRefreshRate, "Process information UI refreshes rate in milliseconds greater than 1000")
```
MadhavJivrajani, line 100:
```suggestion
	procCmd.Flags().Int32VarP("refresh", "r", DefaultProcRefreshRate, "Process information UI refreshes rate in milliseconds greater than 1000")
```

cmd/proc.go
```diff
@@ -71,8 +76,8 @@ Syntax:
```
MadhavJivrajani, line 76:
```suggestion
procRefreshRate, _ := cmd.Flags().GetInt32("refresh")
```

cmd/proc.go
```diff
@@ -43,8 +47,9 @@ Syntax:
   grofer proc -p [PID]`,
 	Aliases: []string{"process", "processess"},
 	RunE: func(cmd *cobra.Command, args []string) error {
-		if len(args) > 0 {
```
MadhavJivrajani, line 46: Why is this check deleted?

cmd/root.go
```diff
@@ -62,6 +73,8 @@ func Execute() {
 func init() {
 	cobra.OnInitialize(initConfig)
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.grofer.yaml)")
+
+	rootCmd.Flags().Int32VarP(&overallRefreshRate, "refresh", "r", DefaultOverallRefreshRate, "Overall stats UI refreshes rate in milliseconds greater than 1000")
```
MadhavJivrajani, line 77:
```suggestion
	rootCmd.Flags().Int32VarP("refresh", "r", DefaultOverallRefreshRate, "Overall stats UI refreshes rate in milliseconds greater than 1000")
```

cmd/root.go
```diff
@@ -20,22 +20,31 @@ import (
...
 	overallGraph "github.com/pesos/grofer/src/display/general"
 	"github.com/pesos/grofer/src/general"
 )
 
+const (
+	DefaultOverallRefreshRate = 1000
+)
+
 var cfgFile string
+var overallRefreshRate int32
```
MadhavJivrajani, line 36:
```suggestion

```

cmd/root.go
```diff
@@ -45,10 +54,12 @@ var rootCmd = &cobra.Command{
```
MadhavJivrajani, line 54:
```suggestion
overallRefreshRate, _ := cmd.Flags().GetInt32("refresh")
```

MadhavJivrajani:
@yongruilin the only reason I've suggested extracting values for flags by doing `cmd.Flags().GetInt32` is for the sake of uniformity

Also, make sure to run your code through `go fmt` before committing!

yongruilin:
@MadhavJivrajani , updated as suggested, please check.

PR #38, opened 2020-08-10
fix proportion of procRefreshRate

### LINKED ISSUE #5: [BUG] High CPU utilisation
**Describe the bug**

`grofer proc` takes anywhere between 40-60% of a CPU core (measured using `grofer proc -p $(pidof grofer)` ). Looking through `htop`, the usage seems even higher, there are multiple grofer proc processes (5+) with two of them taking 60+10% CPU.

`grofer` uses 15-20% of a CPU core (again, measured through `grofer proc -p $(pidof grofer) `). htop reports the CPU usage to be 25%+.

**To Reproduce**

1. Open `grofer proc` in one terminal
2. In another terminal, do `grofer proc -p $(pidof grofer)`
3. Notice high CPU usage

**Expected behavior**

The profiler itself should not consume too much CPU power. Ideally, less than 5% seems good.

**Screenshots**

CPU usage as seen through grofer
CPU usage seen through htop

 - OS: Fedora 32
 - Go: 1.14.6

**Additional context**

Could be solved by reducing the delay/interval to be around 1s.

siddhantrao23:
# Description

Fixes refresh rate sent to `ServeProcs` as 0.8 times `procRefreshRate` for better CPU Performance of `grofer proc` (at 15% now)
Also fixed previous bug of "gg" key event not working and refactored Resizing code to it's event handler for updating UI.

Partly fixes High CPU, more optimizations needed

## Type of change

# Checklist:

### DIFF
--- a/cmd/proc.go
+++ b/cmd/proc.go
@@ -81,7 +81,7 @@ Syntax:
 
 			wg.Add(2)
 
-			go process.ServeProcs(dataChannel, endChannel, int32(2*procRefreshRate/5), &wg)
+			go process.ServeProcs(dataChannel, endChannel, int32(4*procRefreshRate/5), &wg)
 			go procGraph.AllProcVisuals(dataChannel, endChannel, procRefreshRate, &wg)
 			wg.Wait()
 		}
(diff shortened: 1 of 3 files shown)


siddhantrao23:
@Gituser143 Check under the event `<Resize>`, `updateUI()` should only be called on resize instead of for every tick. Unless this doesn't work for you.

PR #53, opened 2020-09-23
Fix error handling under cmd/root

### LINKED ISSUE #48: [BUG] error handling needs to be done correctly
**Describe the bug**
Currently, all errors that might occur are being returned up the function call stack with the intention of being returned to the files in the `cmd` directory. `cobra` provides functionality for handling errors by providing `Run` or `RunE` in the `cmdNameCmd` struct which allows you to return an error to `cobra` and it tries and handles it. 

With the current concurrency model of `grofer`, the functions for retrieving data and rendering data are put in go-routines because of why "returning" an error to that function is rendered pointless.

**Proposal:**
Have a channel of type `error` which blocks waiting for an event to occur, on a successful run, `nil` could be passed to the channel in which case the program can exit, whereas, on the occurrence of an error, the error can be taken out of the channel and handled or returned to another function if need be for handling. 

**To Reproduce**
1. Go to `src/general/cpuInfo.go`
2. Change "mpstat" to "baby-shark-dodododo" (hopefully not a valid command although that would be awesome) on line 52
3. Execute: `go run grofer.go --cpuinfo`
4. Everything renders fine.

**Expected behavior**
Error messages must be displayed. 

 - OS: Kubuntu 20.04
 - Go Version: 1.14.7

**Additional context**
Since such errors aren't handled, the `kill` signal does not execute properly on pressing the `q` key which is meant to exit the program. There is an error that's occurred but that goroutine cannot handle it which is why the `kill` signal works in the rendering part but does not work correctly in the data serving part and it blocks there and the services of `ctrl + c` need to be employed.

egawata:
# Description

This PR fixes the way of error handling under `cmd/root`.

- When any error occurs, it will be returned to callers, which enables to handle it at the right place.
- Pressing `q` terminates execution in a right way.

Fixes #48 

## Type of change

# Checklist:

### DIFF
--- a/src/general/errors.go
+++ b/src/general/errors.go
@@ -1,3 +1,19 @@
+/*
+Copyright © 2020 The PES Open Source Team pesos@pes.edu
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package general
 
 import (
--- a/src/general/errors.go
+++ b/src/general/errors.go
@@ -1,0 +1,7 @@
+package general
+
+import (
+	"errors"
+)
+
+var ErrCanceledByUser = errors.New("canceled by user")
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -16,9 +16,9 @@ limitations under the License.
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"os"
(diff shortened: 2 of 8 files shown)


src/general/errors.go
```diff
@@ -0,0 +1,7 @@
+package general
```
Gituser143, line 17: Hi @egawata since this file is a new addition. We'll need you to put in the license at the top of the file.

Gituser143 (changes requested):

src/general/errors.go
MadhavJivrajani: You can find the license at the beginning of any of the `.go` files

egawata: @Gituser143 
I'm sorry for forgetting that 🙇‍♂️   I've added it at https://github.com/pesos/grofer/pull/53/commits/d48915bd6d30abf0a7aa2e7c2e73a50ddf01bd5c .

MadhavJivrajani (approved):
Great stuff @egawata ! Thanks for your contribution!

Gituser143 (approved):

Issue #42, opened 2020-08-13
[FEATURE REQ] add functionality for getting help for a particular command


MadhavJivrajani:
**Is your feature request related to a problem? Please describe.**
When running a particular interactive command like `grofer` or `grofer proc` it would be nice to have a way to get something like a table of available key bindings and what they do for that command. 

**Describe the solution you'd like**
Pressing a key like `h` should pause the UI rendering and display a table of the available key bindings for that command and what they do like `↑` to scroll, `k` to kill a process in the `grofer proc` command once #16 is done etc.

anihm136:
Can I pick this up?

MadhavJivrajani:
Absolutely!

anihm136:
Awesome! Just to be sure, all the commands are listed in the usage section of the readme right? (i.e, no other commands/flags exist)

MadhavJivrajani:
Yep!
All commands/flags for the latest release are available in the README.

anihm136:
I'm still working on this :smile:

anihm136:
I think I should wait on #54, since my changes will be in the same files

MadhavJivrajani:
Ah okay, cool cool, I'll re-assign you

anihm136:
So what I've been thinking for this is to create a general help widget, which can take a help string and display it as required, as well as handling the necessary keybindings. This can then be used in the 3 subcommands with the appropriate help strings. Do you think this is a good way to go? If so, what would be the best place to define such a widget?

anihm136:
Bump @MadhavJivrajani @Gituser143 @metonymic-smokey

MadhavJivrajani:
Hi @anihm136,
Apologies for the late response, yes that seems like a good way to go about it, just one question:
> which can take a help string and display it as required

I didn't quite get you here, could you give an example about what would count as a "help string"?

Gituser143:
@anihm136 Sorry for the delay! This seems perfect in theory. Maybe use a textbox or table widget to make it work? You could pause the UI rendering and then render just this help widget when the help key is pressed.

If you feel that's doable, you could get started on that!

Gituser143:
> I didn't quite get you here, could you give an example about what would count as a "help string"? 

Something like pressing `h` for help box. That's what was mentioned in the issue description...

MadhavJivrajani:
That's what I understood from the keybindings part.

anihm136:
[...]

PR #54, opened 2020-09-25
Press F to kill

### LINKED ISSUE #16: [FEATURE REQ] Add keybinding to kill processes
Generally I use a system profiler to monitor processes taking up large amounts of system resources, which I would usually need to kill. Having a keybinding to kill a process from the proc page would be very helpful in this regard

A simple single key keybinding to kill a highlighted process from within the proc page

**Describe alternatives you've considered if any**
None

Samyak2:
# Description

Added two keybindings to kill the selected process in `grofer proc`:
  - `K` (shift+K)
  - `<F9>` - because [htop](https://github.com/htop-dev/htop/) uses this keybind

Fixes #16 

## Type of change

# Checklist:

### DIFF
--- a/src/display/process/allProcs.go
+++ b/src/display/process/allProcs.go
@@ -138,10 +138,11 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 
 	previousKey := ""
 	selectedStyle := ui.NewStyle(ui.ColorYellow, ui.ColorClear, ui.ModifierBold)
+	killingStyle := ui.NewStyle(ui.ColorWhite, ui.ColorMagenta, ui.ModifierBold)
+	errorStyle := ui.NewStyle(ui.ColorBlack, ui.ColorRed, ui.ModifierBold)
 
 	// updates process list immediately
 	updateProcs := func() {
-		myPage.BodyList.SelectedRowStyle = selectedStyle
 		if runAllProc {
 			procs, err := proc.Processes()
 			if err == nil {
@@ -165,8 +166,9 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 				updateUI() // updateUI only during resize event
 			case "<Escape>":
 				if paused {
-					pauseProc()
+					runAllProc = true
 					paused = false
+					myPage.BodyList.SelectedRowStyle = selectedStyle
 				}
 			case "s": //s to pause
 				if !paused {
@@ -223,17 +225,21 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 					if !paused {
 						runAllProc = false
 						paused = true
+						myPage.BodyList.SelectedRowStyle = killingStyle
 					} else {
 						// get process and kill it
-						procToKill, err := proc.NewProcess(pid)
+						procToKill, err := proc.NewProcess(pidToKill)
+						myPage.BodyList.SelectedRowStyle = selectedStyle
 						if err == nil {
 							err = procToKill.Kill()
 							if err != nil {
-								return fmt.Errorf("Failed to kill process with PID %d: %v", pid, err)
+								myPage.BodyList.SelectedRowStyle = errorStyle
 							}
+						} else {
+							myPage.BodyList.SelectedRowStyle = errorStyle
 						}
 						runAllProc = true
-						paused = false
+						killSelected = false
 						updateProcs()
 					}
 				}
@@ -247,8 +253,8 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 			}
 
 		case data := <-dataChannel:
-			myPage.BodyList.SelectedRowStyle = selectedStyle
 			if runAllProc {
+				myPage.BodyList.SelectedRowStyle = selectedStyle
 				myPage.BodyList.Rows = getData(data)
 
 				on.Do(updateUI)
@@ -256,6 +262,17 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 
 		case <-tick: // Update page with new values
 			ui.Render(myPage.Grid)
+		}
+
+		myPage.BodyList.SelectedRowStyle = selectedStyle
+		if killSelected {
+			exists, _ := proc.PidExists(pidToKill)
+			if !exists {
+				runAllProc = true
+				killSelected = false
+				myPage.BodyList.SelectedRowStyle = selectedStyle
+				updateProcs()
+			}
 		}
 	}
 }
--- a/src/display/process/allProcs.go
+++ b/src/display/process/allProcs.go
@@ -169,7 +169,9 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 				myPage.BodyList.ScrollTop()
 			case "G", "<End>":
 				myPage.BodyList.ScrollBottom()
-			case "F", "<F9>":
+			case "K", "<F9>":
+				updateUI()
+
 				row := myPage.BodyList.Rows[myPage.BodyList.SelectedRow]
 				// get PID from the data
 				pid64, err := strconv.ParseInt(strings.SplitN(row, " ", 2)[0], 10, 32)
@@ -177,12 +179,15 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 					return fmt.Errorf("Failed to get PID of process: %v", err)
 				}
 				pid := int32(pid64)
+
 				// get process and kill it
 				procToKill, err := proc.NewProcess(pid)
-				if err != nil {
-					return fmt.Errorf("Failed to kill process with PID %d: %v", pid, err)
+				if err == nil {
+					err = procToKill.Kill()
+					if err != nil {
+						return fmt.Errorf("Failed to kill process with PID %d: %v", pid, err)
+					}
 				}
-				procToKill.Kill()
 			}
 
 			ui.Render(myPage.Grid)
--- a/src/display/process/allProcs.go
+++ b/src/display/process/allProcs.go
@@ -32,7 +32,7 @@ import (
 
 var runAllProc = true
 
-func getData(procs []*proc.Process) []string {
+func getData(procs []*proc.Process) ([]string) {
 	var data []string
 	for _, info := range procs {
 		exe, err := info.Exe()
@@ -170,6 +170,14 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 				myPage.BodyList.ScrollTop()
 			case "G", "<End>":
 				myPage.BodyList.ScrollBottom()
+			case "F", "<F9>":
+				row := myPage.BodyList.Rows[myPage.BodyList.SelectedRow]
+				// get PID from the data
+				pid64, _ := strconv.ParseInt(strings.SplitN(row, " ", 2)[0], 10, 32)
+				pid := int32(pid64)
+				// get process and kill it
+				procToKill, _ := proc.NewProcess(pid)
+				procToKill.Kill()
 			}
 
 			ui.Render(myPage.Grid)
--- a/src/display/process/allProcs.go
+++ b/src/display/process/allProcs.go
@@ -17,22 +17,23 @@ limitations under the License.
 package process
 
 import (
+	"context"
 	"fmt"
 	"log"
-	"os"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
 
 	ui "github.com/gizak/termui/v3"
+	info "github.com/pesos/grofer/src/general"
 	"github.com/pesos/grofer/src/utils"
 	proc "github.com/shirou/gopsutil/process"
 )
 
 var runAllProc = true
 
-func getData(procs []*proc.Process) ([]string) {
+func getData(procs []*proc.Process) []string {
 	var data []string
 	for _, info := range procs {
 		exe, err := info.Exe()
@@ -106,13 +107,14 @@ func getData(procs []*proc.Process) ([]string) {
 }
 
 func AllProcVisuals(dataChannel chan []*proc.Process,
-	endChannel chan os.Signal,
-	refreshRate uint64,
-	wg *sync.WaitGroup) {
+	ctx context.Context,
+	refreshRate int32) error {
 
 	if err := ui.Init(); err != nil {
 		log.Fatalf("failed to initialize termui: %v", err)
 	}
+
+	defer ui.Close()
 
 	var on sync.Once
 
@@ -142,10 +144,7 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 		case e := <-uiEvents:
 			switch e.ID {
 			case "q", "<C-c>": //q or Ctrl-C to quit
-				endChannel <- os.Kill
-				ui.Close()
-				wg.Done()
-				return
+				return info.ErrCanceledByUser
 			case "<Resize>":
 				updateUI() // updateUI only during resize event
 			case "s": //s to pause
@@ -173,10 +172,16 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 			case "F", "<F9>":
 				row := myPage.BodyList.Rows[myPage.BodyList.SelectedRow]
 				// get PID from the data
-				pid64, _ := strconv.ParseInt(strings.SplitN(row, " ", 2)[0], 10, 32)
+				pid64, err := strconv.ParseInt(strings.SplitN(row, " ", 2)[0], 10, 32)
+				if err != nil {
+					return fmt.Errorf("Failed to get PID of process: %v", err)
+				}
 				pid := int32(pid64)
 				// get process and kill it
-				procToKill, _ := proc.NewProcess(pid)
+				procToKill, err := proc.NewProcess(pid)
+				if err != nil {
+					return fmt.Errorf("Failed to kill process with PID %d: %v", pid, err)
+				}
 				procToKill.Kill()
 			}
 
--- a/src/display/process/allProcs.go
+++ b/src/display/process/allProcs.go
@@ -129,7 +129,7 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 
 	updateUI() // Render empty UI
 
-	pause := func() {
+	pauseProc := func() {
 		runAllProc = !runAllProc
 	}
 
@@ -139,6 +139,22 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 	previousKey := ""
 	selectedStyle := ui.NewStyle(ui.ColorYellow, ui.ColorClear, ui.ModifierBold)
 
+	// updates process list immediately
+	updateProcs := func() {
+		myPage.BodyList.SelectedRowStyle = selectedStyle
+		if runAllProc {
+			procs, err := proc.Processes()
+			if err == nil {
+				myPage.BodyList.Rows = getData(procs)
+
+				on.Do(updateUI)
+			}
+		}
+	}
+
+	// whether the actual UI controls are paused
+	paused := false
+
 	for {
 		select {
 		case e := <-uiEvents:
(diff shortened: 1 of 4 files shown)


MadhavJivrajani (approved):
Looks good!
PS love the PR title xD

Gituser143 (commented):
@Samyak2 is it necessary to get the `info` for the list of processes everytime data is sent through the channel? Would it be possible to maybe fetch this only when we need to kill the process? 

PS: This is just a thought, might not work though :P

MadhavJivrajani:
That's interesting, maybe you could do something like as follows,
You could get the PID of the process as the first element of the space-separated string: 
```
# not go code
pid = myPage.BodyList.Rows[myPage.BodyList.SelectedRow].Split(...)
```

After getting the PID, you could get the process type from `gopsutil` by calling the [NewProcess](https://godoc.org/github.com/shirou/gopsutil/process#NewProcess) function:
```go
procToKill, err := NewProcess(pid)
```
And on this `process` type you could call the `Kill` function as you already have.

We will have to implement the error handling workflow as implemented in #53 , this can come under a separate PR.

metonymic-smokey (approved):

Gituser143:
@MadhavJivrajani That sounds good, @Samyak2 what do you think?

Samyak2:
Sounds good. I thought we would need the `procs` list for some other operations, but since we'll need details of only one process Madhav's suggestion is a better idea. I'll make the changes.

Samyak2:
I have added some error handling, but I'm not sure if that's the right way to do it.

MadhavJivrajani (commented):
The changes to execute the `Kill()` function look good. However, the error handling part might work in the sense that it will terminate the execution of the program, but there the error won't be propagated up the function call lineage in order for it to be visible/gracefully terminate. If you're interested in working on the error handling part of this as well then we'd be more than happy to help you through it!

Samyak2:
Yes! I would like to work on it. I'm looking at #53 to get some idea.

I'll also fix the formatting issue.

MadhavJivrajani:
Awesome!
Let us know if you need any help :smile:

Samyak2:
I noticed that there's no error handling in `cmd/proc.go`, so I'll have to add it similar to #53 with the following changes:
 - Make an `errgroup`, call the goroutines through the error group and then collect the error and print.
 - Change the `process` and `procGraph` functions to  use `context.Content` instead of `endChannel` and to return `error`.

Is this all there is to it or am I missing something?

MadhavJivrajani:
That sounds about right. 
There is [another place](https://github.com/pesos/grofer/blob/c10ba49a1b8d5dd01418ef77f0b68755971690c6/src/display/process/procGraphs.go#L57) (unrelated to this PR) where error handling needs to be implemented, you can take it up along with this PR itself or not, either way, your approach LGTM.

Samyak2:
I have implemented error handling as in #53, but I've run into an issue. When an error is encountered (or when q is pressed), grofer exits and shows the error message but the UI does not seem to exit gracefully - cursor is invisible in the prompt and the UI remains in the terminal output.

image

MadhavJivrajani:
Hi,
I was able to figure out the issue with that, it seems to be an error from our end. For some reason, for the function `AllProcVisuals()` a `defer ui.Close()` seemed to be missing because of which the graceful exit wasn't happening. The reason it was working so far was that we were issuing an `os.Kill` whenever the events `q` or `<C c>` took place. Placing a `defer ui.Close()` after line 133 in your current code should do the trick! My apologies for the confusion.

Samyak2:
That works! Thanks :smile:

Samyak2:
I have a concern wrt this PR. The shortcut "F" is a little too close to "G" (for scrolling to the bottom), it might lead to accidentally killing a process (I think this happened to me twice, had to restart/relogin because the window manager stopped working).

I was thinking of changing it to "K" (or any other alternative that might better). What do you think?

(RIP title though)

MadhavJivrajani:
Ah damn 
That makes sense, yeah that sounds good go ahead with `K` itself.

Samyak2:
I ran into another issue :sweat:. Seems to be some kind of race condition.
To reproduce:
 - Start grofer proc with a high refresh rate. `./grofer proc -r 5000`
 - Start a non-terminating process in another terminal: Ex: `cat`
 - Go to the newly created `cat` process in grofer
 - Terminate (ctrl-c) the `cat` process
 - Wait a second or two and try to kill the cat process ('F')
 - See this error (repeat if it says "process does not exist")

image

It seems that the `SelectedRow` is not updated when the list is updated. This can sometimes lead to the wrong process being killed. For example, repeat the above steps but run `cat` before running grofer and it will lead to `grofer` getting killed instead of cat (assuming no other processes were created during this time)

image

Gituser143:
@Samyak2 `process.Kill()` returns an error type, maybe we could use this in anyway? Apart from this, a simple call to `updateUI()` at the beginning of the case statement to kill might just help.

Gituser143:
@Samyak2 And as for the second issue when `cat` is killed externally first and then attempted to be killed through `grofer`, a naive approach which I feel might work is something like this:

```go
case "F", "<F9>":
  initialRow := myPage.BodyList.Rows[myPage.BodyList.SelectedRow]
  // get initial PID from the data
  initialPid64, err := strconv.ParseInt(strings.SplitN(initialRow, " ", 2)[0], 10, 32)
  // handle err
  initialPid := int32(initialPid64)
  updateUI()
  // get row selected by grofer
  row := myPage.BodyList.Rows[myPage.BodyList.SelectedRow]
  // get PID from the data
  pid64, err := strconv.ParseInt(strings.SplitN(row, " ", 2)[0], 10, 32)
  pid := int32(pid64)
  // get process and kill it
  procToKill, err := proc.NewProcess(pid)
  if err == nil && pid == initialPid {
    procToKill.Kill()
  }
```

Samyak2:
> @Samyak2 `process.Kill()` returns an error type, maybe we could use this in some way?

Oh yes, I must have missed that while doing the error handling.

@Gituser143 I tried your solution, it did not fix either of the issues - index of out range and the issue of killing the wrong process.

```go
case "K", "<F9>":
	initialRow := myPage.BodyList.Rows[myPage.BodyList.SelectedRow]
	// get initial PID from the data
	initialPid64, err := strconv.ParseInt(strings.SplitN(initialRow, " ", 2)[0], 10, 32)
	if err != nil {
		return fmt.Errorf("Failed to get PID of process: %v", err)
	}
	initialPid := int32(initialPid64)

	updateUI()

	row := myPage.BodyList.Rows[myPage.BodyList.SelectedRow]
	// get PID from the data
	pid64, err := strconv.ParseInt(strings.SplitN(row, " ", 2)[0], 10, 32)
	if err != nil {
		return fmt.Errorf("Failed to get PID of process: %v", err)
	}
	pid := int32(pid64)

	// get process and kill it
	procToKill, err := proc.NewProcess(pid)
	if err != nil {
		return fmt.Errorf("Failed to get process with PID %d: %v", pid, err)
	} else if pid == initialPid {
		err = procToKill.Kill()
		if err != nil {
			return fmt.Errorf("Failed to kill process with PID %d: %v", pid, err)
		}
	}
```

Gituser143:
@Samyak2 I tried this out in a similar but sloppy way. Here you're checking if `err != nil` and if so returning/printing error. Instead, avoid that case entirely and only work on cases if `err == nil`. 

```go
case "F", "<F9>":
				initialRow := myPage.BodyList.Rows[myPage.BodyList.SelectedRow]
				// get PID from the data
				initialPid64, err := strconv.ParseInt(strings.SplitN(initialRow, " ", 2)[0], 10, 32)
				if err != nil {
					return fmt.Errorf("Failed to get PID of process: %v", err)
				}
				initialPid := int32(initialPid64)
				updateUI()
				row := myPage.BodyList.Rows[myPage.BodyList.SelectedRow]
				// get PID from the data
				pid64, err := strconv.ParseInt(strings.SplitN(row, " ", 2)[0], 10, 32)
				if err != nil {
					return fmt.Errorf("Failed to get PID of process: %v", err)
				}
				pid := int32(pid64)
				// get process and kill it
				procToKill, err := proc.NewProcess(pid)
				if err == nil && pid == initialPid {
					procToKill.Kill()
				}
			} 
```

Samyak2:
> Here you're checking if `err != nil` and if so returning/printing error. Instead, avoid that case entirely and only work on cases if `err == nil`.

@Gituser143 Ah right, I missed that.

That does help when the process has already exited, but if the row list is updated without updating the `SelectedRow` it still fails with a "index out of range" panic (even with `UpdateUI`)

Gituser143:
@Samyak2 the confirmation sounds like a good idea, I hadn't brought that up previously because I'm not sure if we can do that without creating another widget entirely. If you feel this can be done without overly complicating it, I'm all for it.

PS: could you tell me how to replicate the index out of range error? I can't seem to be able to do so.

Samyak2:
I'm not familiar with the UI library, so I'm not sure if I can do it quickly. (Perhaps we could open a new issue for this?)

To reproduce the index out of range error (from the commit I just pushed):
 - `./grofer proc -r 5000`
 - Run `cat` on another terminal
 - Find the `cat` process in grofer
 - Ctrl-C the `cat` process on the other terminal
 - In grofer: Spam the kill key ('K') on the `cat` process until it exits with a panic.

Gituser143:
@Samyak2 I just looked at the implementation of kill in gotop. Here, exec is used to kill using the kill command itself. Might or might not be helpful :P

Gituser143:
> I'm not familiar with the UI library, so I'm not sure if I can do it quickly. (Perhaps we could open a new issue for this?)
> 
> To reproduce the index out of range error (from the commit I just pushed):
> 
>     * `./grofer proc -r 5000`
> 
>     * Run `cat` on another terminal
> 
>     * Find the `cat` process in grofer
> 
>     * Ctrl-C the `cat` process on the other terminal
> 
>     * In grofer: Spam the kill key ('K') on the `cat` process until it exits with a panic.

Also, if we update the UI right after the kill would it help in anyway? Although naive, I feel that just might help.

Gituser143:
@Samyak2 I just realised, maybe we could use a waitGroup to ensure nothing else happens while killing. This way even if keys are spammed after the first F9, it won't even be recorded. What do you think?

PS: I think 'k' for kill isn't appropriate as it's being used currently to scroll. I only realised this, when I couldn't kill procs now

Samyak2:
> @Samyak2 I just realised, maybe we could use a waitGroup to ensure nothing else happens while killing. This way even if keys are spammed after the first F9, it won't even be recorded. What do you think?

I'm not entirely sure how a wait group works yet, but is your suggestion is to do `wg.Wait` before killing?

> PS: I think 'k' for kill isn't appropriate as it's being used currently to scroll. I only realised this, when I couldn't kill procs now 

The key for killing is `K` (shift K) and not `k` 😄

Gituser143:
> The key for killing is `K` (shift K) and not `k` smile
I completely missed that, my bad. 

> I'm not entirely sure how a wait group works yet, but is your suggestion is to do `wg.Wait` before killing?

We wait while it's killing. A waitGroup is essentially a semaphore. We set it's value to 1 (`wg.Add(1)`), call a routine to kill the process and `wg.Wait()` till that routine is finished (call `wg.Done()` inside the routine). This might help avoiding key spam, but it certainly won't fix the index out of range issue (when we try to get the selected row, but the row was just removed due to the process exiting).

MadhavJivrajani:
Calling `updateUI()` won't solve the issue here since it's only resizing the widgets. IMO you will have to call the [`Processes()`](https://godoc.org/github.com/shirou/gopsutil/process#Processes) function as done [here](https://github.com/pesos/grofer/blob/c10ba49a1b8d5dd01418ef77f0b68755971690c6/src/process/serveData.go#L51) and then update the rows as done in line 220 of your current code. 

A cleaner way to check if a process exists when you're calling the `Kill()` function would be to use the [`PidExists()`](https://godoc.org/github.com/shirou/gopsutil/process#PidExists) function and handle the error in a similar way as done so far. Only if the PID exists then kill the process, if not then don't do anything, and only after doing this, call the `Processes()` function and update the list of processes so that it doesn't have to wait till the next tick.

Samyak2:
>  IMO you will have to call the [`Processes()`](https://godoc.org/github.com/shirou/gopsutil/process#Processes) function as done [here](https://github.com/pesos/grofer/blob/c10ba49a1b8d5dd01418ef77f0b68755971690c6/src/process/serveData.go#L51) and then update the rows as done in line 220 of your current code.

The UI updating as soon as you try to kill a process does not seem like the best solution, but I guess it's better than waiting for the next UI update and accidentally killing the wrong process.

> A cleaner way to check if a process exists when you're calling the `Kill()` function would be to use the [`PidExists()`](https://godoc.org/github.com/shirou/gopsutil/process#PidExists) function and handle the error in a similar way as done so far. 

This is a better solution than just ignoring the error from `Kill()` :smile: 

> Only if the PID exists then kill the process, if not then don't do anything, and only after doing this, call the `Processes()` function and update the list of processes so that it doesn't have to wait till the next tick.

With this, the issue of killing the wrong process still exists. I think it would be better to update the UI after K is pressed before doing anything else.

(Sorry for the late response :disappointed:)

MadhavJivrajani:
> With this, the issue of killing the wrong process still exists. I think it would be better to update the UI after K is pressed before doing anything else.
Yes, you're right, my bad.

Gituser143:
@Samyak2 An approach we could take would be to get the row as soon as F is pressed, we can then extract the PID from that row. Once this happens, we can call the `pause()` function to stop updating UI. 

Once we get the PID from that row, we can confirm with the user if the process they want to kill matches the one we caught. Followed by a (y/n) prompt through a text box widget. 

On confirmation, we can check if the process still exists and kill it.

What are your thoughts on this?

MadhavJivrajani:
Seems forced tbf, it shouldn't be more work than it has to be

Samyak2:
@Gituser143 I like the idea. A confirmation is, IMO, quite important for this feature since it's `SIGKILL`ing a process.

`htop` does it in a similar way except that the selection is moved as the list is getting updated.

MadhavJivrajani:
Confirmation of some form is important, agreed, but if the same issue as before occurs, that is if the wrong process is selected for killing, then prompting that and the user selecting `no` and then repeating it, that might not be what should be done to guarantee the best experience, and if the same thing happens more than twice it could also count as borderline annoying for the user.

I might be misunderstanding the whole thing, please correct me if I've gotten it wrong.

Samyak2:
That is true, it will get annoying if that happens too many times in a row, but I don't think there's a simple solution that can fix this. For context, `htop` (i'm sorry that's the only other profiler i know :laughing:) behaves similarly - the process list changes quickly and a wrong process might get selected but since there is a confirmation (selection of signal to send), the wrong process is never killed.

For now, I was thinking of doing the following:
 - Update the process list when `K` is pressed
 - If the PID exists, display a confirmation
 - If confirmed, kill if the PID still exists

(Also, another thing I noticed in `htop` was that, when a process is selected and it dies while it was selected, the selection is automatically cleared)

MadhavJivrajani:
I see what you mean, `htop` also provides a list of signals that could help kill the process. That's what should be done as well instead of directly send a `SIGKILL`. Afaik `gopsutil` does not provide functionality for other signals other than `SIGKILL`, but implementing that is straightforward. If we can decide on this approach of killing a process (i.e. providing signal options to the users), then I can open a separate issue for implementing kill signal functionality and this PR can be labeled as long term.

Gituser143:
> Confirmation of some form is important, agreed, but if the same issue as before occurs, that is if the wrong process is selected for killing, then prompting that and the user selecting no and then repeating it, that might not be what should be done to guarantee the best experience, and if the same thing happens more than twice it could also count as borderline annoying for the user.

This won't be an issue. Because as soon as we get the row, we can kill slowly and if it doesn't exist, return gracefully. Since we update UI right after this above return/process killing. The user won't be clicking wrong PIDs again and again

Samyak2:
> I see what you mean, `htop` also provides a list of signals that could help kill the process. That's what should be done as well instead of directly send a `SIGKILL`. Afaik `gopsutil` does not provide functionality for other signals other than `SIGKILL`, but implementing that is straightforward.

According to the docs, `gopsutil` only supports 4 types of signals being sent to a process ([source](https://godoc.org/github.com/shirou/gopsutil/process#Process.SendSignal)). Is there another way of sending a signal to a process?

> If we can decide on this approach of killing a process (i.e. providing signal options to the users), then I can open a separate issue for implementing kill signal functionality and this PR can be labeled as long-term.

How about this: for this PR, I'll implement a simple confirmation - press K once to select a process, pause UI, press K again to kill (if it still exists).
You can then open an issue to extend this to any signal instead of just `SIGKILL` where we'll need a separate UI component with the list of signals.

Samyak2:
@Gituser143 @MadhavJivrajani what do you think of the changes? Is it a good way to do this?

(also, the `hacktoberfest-accepted` label isn't required, this PR was opened before October)

src/display/process/allProcs.go
```diff
@@ -155,6 +155,7 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 
 	previousKey := ""
 	selectedStyle := ui.NewStyle(ui.ColorYellow, ui.ColorClear, ui.ModifierBold)
+	killingStyle := ui.NewStyle(ui.ColorWhite, ui.ColorRed, ui.ModifierBold)
 
 	// updates process list immediately
 	updateProcs := func() {
```
Gituser143, line 161: @Samyak2 In this function since you're using the `sync.Once` variable, will it really call `updateUI`? Because the once variable is used to render the very first data coming in from the data channel.

Gituser143: Apart from this, the `updateProcs()` function is called at the very beginning of the kill section when `K` is pressed. I'm not sure what your reasoning is to call that there. Please let me know?

Samyak2: @Gituser143 you are right, it won't actually call `updateUI`. I had copied this part from `case data := <-dataChannel:` and forgot to remove the Once line 😅.
The reasoning behind calling `updateProcs` at the beginning of the kill section was to ensure that the process selected for kill actually exists. But upon further testing, I found that this leads to a different process getting selected for killing.

IMO a better approach would be doing this:
 - After a process is selected for killing, monitor it in every update
 - if the PID still exists, do nothing
 - if the PID does not exist, remove selection

src/display/process/allProcs.go
```diff
@@ -278,5 +275,15 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
...
+		if killSelected {
+			exists, _ := proc.PidExists(pidToKill)
+			if !exists {
+				runAllProc = true
+				killSelected = false
+				myPage.BodyList.SelectedRowStyle = selectedStyle
+				updateProcs()
+			}
+		}
 	}
```
Gituser143, line 288: @Samyak2 This method of checking is accurate but the issue is we'll never get to this section of code as the above statements are in an infinite loop. The only time we exit from that loop is when we exit the program entirely! What you're doing in the else clause of `if !killSelected` suffices I feel.

src/display/process/allProcs.go
```diff
@@ -182,78 +181,76 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
...
 						// get process and kill it
-						procToKill, err := proc.NewProcess(pid)
+						procToKill, err := proc.NewProcess(pidToKill)
 						if err == nil {
 							err = procToKill.Kill()
 							if err != nil {
-								return fmt.Errorf("Failed to kill process with PID %d: %v", pid, err)
+								return fmt.Errorf("Failed to kill process with PID %d: %v", pidToKill, err)
 							}
 						}
```
Gituser143: @Samyak2 Wouldn't returning this error hinder with the UI? Because in the `proc.Kill()` function, it checks again if the PID exists and then sends a signal. I'm against returning an error here if `proc.kill()` fails as it would unnecessarily stop the UI and quit. I'd drop the check entirely.

Maybe let's wait for @MadhavJivrajani for a better opinion?

Gituser143 (changes requested):

src/display/process/allProcs.go
Samyak2: This is inside the loop xD, right after the `select`

src/display/process/allProcs.go
Samyak2: I'm returning errors in two places (inside kill):
 - If parsing the PID fails: this shouldn't happen in any condition, unless there's a non-numeric PID printed somehow. I think we should keep this since this error will mean that something is seriously wrong.
 - Process kill fails: this could mean permission issues. We could probably ignore this, but there must be some way to let the user know that it failed.

src/display/process/allProcs.go
Gituser143: Oops, my bad!

src/display/process/allProcs.go
Gituser143: @Samyak2 

> but there must be some way to let the user know that it failed.

I'd prefer this than closing the UI. Maybe render a temporary textbox to deliver the message?

MadhavJivrajani: Yes, I'd also prefer some way of letting the user know and not closing the UI. Maybe instead of the textbox, we could have something simpler like making the color of the selected row as red or something of the sort and we could mention the meaning of doing this as part of #42

Gituser143: I thought of the same thing. But I thought it'll hinder experience. Considering the user will have to keep popping the help text open just to understand what's going on. Maybe we could get the row to turn red for now and maybe open another issue to address the delivery of the message?

Samyak2: Yes, sounds good. Currently, the row is highlighted in red when it is selected for killing, I'll change the colors.

Gituser143:
@Samyak2 On pressing `K` for the first time, if we wait for more than `refreshRate` amount of time, the selection is unhilighted. Could you please look into this?

Samyak2:
Oops

Gituser143 (approved):

Gituser143:
@Samyak2 This looks good! There are a few conflicts that have to be resolved though. Could you please look into that?
Conflicting files are:
* cmd/proc.go
* src/display/process/allProcs.go
* src/display/process/procGraphs.go
* src/process/serveData.go

Samyak2:
Yes, I'll rebase on master

Samyak2:
^ Rebased on upstream/master and squashed a few commits

Gituser143 (approved):

Gituser143:
@Samyak2 Everything looks good! Great work! I'll wait for @MadhavJivrajani or @metonymic-smokey to merge if they have no issues.

Gituser143:
@Samyak2 There seems to be 2 issues with `grofer proc -p PID`.

1. It's missing a `ui.Render(myPage.grid)` at the end of the cases. This causes a significant lag while scrolling through the child process list. Refer to #60 to see what's missed.

2. When trying to exit from the UI, it doesn't close cleanly. There's a signifiant delay there too.

Gituser143:
@Samyak2 there seems to be some issue with `grofer proc -p PID`. The UI doesn't exit properly. Could you please look into this?

Samyak2:
Right, the UI was not being closed in `ProcVisuals`.

Also, I noticed that it takes a second or two to exit. Do you know why?

Gituser143:
> Also, I noticed that it takes a second or two to exit. Do you know why?

I was looking at it. I wasn't able to figure that out. I'll keep at it. I'll update you if I figure something out.

Gituser143 (approved):

Gituser143:
@MadhavJivrajani @metonymic-smokey Plis review once if there's anything missing and merge.

@Samyak2 this is great work! Thanks for the PR!.
PS: You could raise a seperate issue for the UI closing delay.

MadhavJivrajani (approved):
@Samyak2 thanks for your contribution!
It was definitely a very interesting conversation that evolved during the course of the PR :smile: 
@Gituser143 LGTM, merge at your leisure :+1:

Samyak2:
> It was definitely a very interesting conversation that evolved during the course of the PR

:100: this PR was a journey

PR #57, opened 2020-09-25
Add Resizing feature to the memory graphs

### LINKED ISSUE #55: [BUG] Resizing issue with the Memory graphs
**Describe the bug**
The Graphs under the memory section don't resize along with the window size

**To Reproduce**
1. Run ./grofre
2. Resize the window

**Expected behavior**
To be able to see the graphs and the different memory elements' values even on a smaller window

**Screenshots**
Screenshot from 2020-09-25 10-31-53
Screenshot from 2020-09-25 10-31-59

 - OS: Ununtu 20.04
 - Go Version: 1.14.1

nsankethreddy:
# Description

Added Resizing Feature to the memory graphs

Fixes #55 

## Type of change

# Checklist:

# Results
Screenshot from 2020-09-25 11-20-09

Screenshot from 2020-09-25 11-20-17

### DIFF
--- a/src/display/general/overallGraphs.go
+++ b/src/display/general/overallGraphs.go
@@ -187,6 +187,18 @@ func RenderCharts(ctx context.Context,
 					height := int(h / numCores)
 					heightOffset := h - (height * numCores)
 
+					// Get the Bar width
+					myPage.MemoryChart.BarWidth = (w / 16)
+
+					if myPage.MemoryChart.BarWidth > 8 {
+						myPage.MemoryChart.BarWidth = 8
+					}
+
+					if myPage.MemoryChart.BarWidth < 6 {
+						myPage.MemoryChart.BarWidth = 4
+						myPage.MemoryChart.NumStyles = []ui.Style{ui.NewStyle(ui.ColorWhite)}
+						myPage.MemoryChart.BarColors = []ui.Color{ui.ColorBlack, ui.ColorBlack}
+					}
 					// Adjust Memory Bar graph values
 					myPage.MemoryChart.BarGap = ((w / 2) - (4 * myPage.MemoryChart.BarWidth)) / 4
 


Gituser143:
@nsankethreddy currently memory graphs are made with dynamic bar gaps and fixed bar width. These are changed in the `updateUI()` function in `overallGraphs.go`. The bar width is fixed so that the graph doesn't have L-shaped bars as you've shown above. 

Resizing of the graph does happen according to terminal widths, although if terminal is too small, values then exceed the terminal. If you would want to make the bar widths dynamic too, you might want to change them accordingly in the said `updateUI()` function and not in the initialization in `init.go`.

Gituser143:
@nsankethreddy although do keep in mind that when resizing we might have to adhere to a minimum bar width to avoid these L-shaped graphs. I hope I was clear with the message above and do let us know if this (making dynamic bar width) was your intent!

Gituser143 (changes requested):
@nsankethreddy you just removed what made our bar gaps dynamic? I'm unsure what you're trying here?

nsankethreddy:
> @nsankethreddy although do keep in mind that when resizing we might have to adhere to a minimum bar width to avoid these L-shaped graphs. I hope I was clear with the message above and do let us know if this (making dynamic bar width) was your intent!

Yes the intent was to make the bar width dynamic

nsankethreddy:
> @nsankethreddy you just removed what made our bar gaps dynamic? I'm unsure what you're trying here?

@Gituser143 I messed up a lil, still trying to figure it out.

nsankethreddy:
Changes made
- If the width is lesser than a particular value, then do not render the ugly overlapping graphs instead just print the respective values
- If the length is greater than a particular value then fix the bar width size(to prevent the extremely wide graphs)
- Dynamically change the bar width size based on the window size

These are the changes made in this commit

ps: Sorry for the galeej commit message
Screenshot from 2020-09-25 13-31-51
Screenshot from 2020-09-25 13-31-51

Gituser143:
@nsankethreddy You're not removing the graph, instead just making it's colour black. This is specific to terminal colours. 
image

As you can see, in my case it'll end up looking like that. I don't think removing the graph is a good idea. Overall here are a few things you might need to change:
* Currently whatever change you've made is in the `sync.Once` function and not the `updateUI` function. So this won't handle UI being updated.
* We don't want the graph to be removed in case of smaller terminals. Instead set a minimum and maximum bar width and set the dimensions of the bar width in proportion to the terminal dimensions, like how that of bar gap is done. It's alright if all content is not seen in a small terminal.
* Do not attempt to make the bar invisible. That doesn't work well. But if you're insistent you could try replacing it with a widget that is more compact and fits in small spaces.

nsankethreddy:
This is kinda already solved, I don't think the current implementation can be bettered.

Issue #33, opened 2020-08-07
[FEATURE REQ] Add functionality to export data to JSON/CSV.


Gituser143:
**Is your feature request related to a problem? Please describe.**
As of now we only display stats through the UI, exporting these to a file in the format of JSON, CSV, etc., will be extremely useful.

**Describe the solution you'd like**
A functionality where running `grofer export [ --type=JSON ] [ --iter=100 ] [ --interval=500 ] FILE` would export stats over a specified number of iterations over given intervals into a specific file format.

MadhavJivrajani:
I've created a project board for this feature, feel free to close this issue and create and add issues related to the `export` command to the `grofer export` project board!

MadhavJivrajani:
Development for `export` feature is being done on the `export` branch ([link](https://github.com/pesos/grofer/tree/export))

PR #63, opened 2020-10-03
[FEATURE] JSON export functionality

### LINKED ISSUE #34: [FEATURE REQ] add flags to the `export` command
Flags for `refresh` (refresh rate), `iterations` and `type` should be added to `export` command.

In `cmd/export.go` non-persistent flags for `refresh`, `iterations` and `type` should be added and used similar to how refresh rate and pid flags are added to `cmd/proc.go`.

MadhavJivrajani:
# Addition of `export` feature for JSON.

- Adds functionality for exporting profiling information in JSON format.
- User-defined parameters include:
    - Export type (json, others TODO)(default: `json`)
    - Refresh rate in milliseconds (default: `1000 ms`)
    - ~File name (default: `grofer_profile.<export type>`)~
    - File name (default: `grofer_profile`)
    - Iterations (default: `10`)
- JSON is directly written using `io.Writer` and is not marshalled, done to save intermediate memory usage.

Fixes #34 
Partially solves #33 
## Type of change

# Checklist:


cmd/export.go
```diff
@@ -16,24 +16,101 @@ limitations under the License.
...
 )
 
+const (
+	DefaultExportRefreshRate = 1000
+	DefaultExportIterations  = 10
+	DefaultExportFileName    = "grofer_profile"
+	DefaultExportType        = "json"
+)
+
+func validateFileName(fileName string, exportType string) error {
```
Gituser143, line 33: @MadhavJivrajani Do you think enforcing file names is a good idea? I feel would be better if the user could name the exported file whatever he/she pleases.

src/export/general/export.go
```diff
@@ -124,33 +132,42 @@ func (data *OverallStats) updateData() error {
...
-func getJSONData(iter int64, interval uint64) ([]OverallStats, error) {
+func getJSONData(iter uint32, refreshRate uint64) ([]OverallStats, error) {
 	var data []OverallStats
-	var i int64
-	stats := &OverallStats{}
+	var i uint32
+	stats := NewOverallStats()
 	for i = 0; i < iter; i++ {
 		err := stats.updateData()
 		if err != nil {
```
Gituser143, line 155: Here again, since you're not using `i` anywhere except the for loop, might as well use `for i := 0; i < iter; i++`? Reduces number of lines of code is all.

Gituser143 (changes requested):

cmd/export.go
MadhavJivrajani: The user can name it as they please. The constants are to handle un-specified parameters.

src/export/general/export.go
MadhavJivrajani: `i:=0` will infer the type of `i` as `int` and comparison won't be valid with `uint32`. Maybe I can declare it in the loop.

Gituser143: Ah, I missed that. Change if you want to, this works too.

MadhavJivrajani: @Gituser143 turns out I can't do [that](https://stackoverflow.com/questions/43567914/how-to-declare-variable-types-for-loop-variables-in-go/43567978).

cmd/export.go
Gituser143: Say in the case when a user provides `myprofile` and `json` as filenames and export types, the if condition on line 70 would fail and this will then cause the `validateFileName()` function to fail too.

I felt we could drop the appending of extension entirely as this might cause issues when users attempt to script with grofer and the filenames change. But if necessary to append a file extension, we could probably check for the export type as a suffix and act on that accordingly.

What do you think?

src/export/general/export.go
Gituser143: How about `for i := uint32(0); i < iter; i++` :P

cmd/export.go
MadhavJivrajani: The if condition on line 70 should fail right, we want a filename with an extension. The appending is done only in case of default file name (the case where the user does not specify a file name). The appending is not done always.

> But if necessary to append a file extension, we could probably check for the export type as a suffix and act on that accordingly.

That's what's being done. The suffix (last element of the `split` slice (split based on `.`)) is validated, including the case of the user not specifying the extension at all.

> as this might cause issues when users attempt to script with grofer and the filenames change.

Could you give an example of this? Not too sure about what you mean.

Gituser143: > Could you give an example of this? Not too sure about what you mean.

Consider this shell script.

```sh
# You can ignore the below command execution, just need the var.
FILENAME=$(sed 's/old_text/new_text/g' some_file | grep "some pattern") 
grofer export --type json --filename $FILENAME
cat $FILENAME | grep "something"
```

In the current case, the user would probably have to use `cat ${FILENAME}.json` in the last line

> That's what's being done. The suffix (last element of the split slice (split based on .)) is validated, including the case of the user not specifying the extension at all.

How are you handling this? If no extension is given, the filename will be sent to `validateFileName()` and there the check will fail right?

MadhavJivrajani: Wait so, you want it to allow filenames that don't have extensions? 
I wrote the checks to enforce that every file provided (except the one that is unspecified) has an extension and a valid one (corresponding to the passed export type arg). IMO it would be better to let the user handle doing `cat ${FILENAME}.json` rather than creating a file with no extension.

MadhavJivrajani: Hmm actually I see your point now, I guess it would be better to leave providing an extension for the file to the user rather than enforcing it.

Gituser143: > Wait so, you want it to allow filenames that don't have extensions?

Yes, I figured it would be better to print it out exactly as requested.

Gituser143: If adamant about appending file name, maybe display a message saying "Succesfuly exported to file.json" maybe?

MadhavJivrajani: @Gituser143 I've made a few changes. Let me know what you think. I've retained appending an explicit type in the default case since if a file name is not specified then a valid extension being present in the filename would ensure lack of any ambiguity.

MadhavJivrajani: @Gituser143 I've made a few changes. Let me know what you think.

MadhavJivrajani: > If adamant about appending file name, maybe display a message saying "Succesfuly exported to file.json" maybe?

I've removed that aspect completely In the current commit

cmd/export.go
```diff
@@ -30,18 +30,48 @@ const (
...
 
+	// If the file extension is something that grofer does not recognise
+	// then it assumes that it is a valid type and trusts the user on the sme.
 	return nil
 }
 
+func validateFileName(fileName, exportType string) error {
+	isValid := hasValidExtension(fileName, exportType)
+	return isValid
+}
```
Gituser143, line 73: @MadhavJivrajani These two functions have the same signature, why not directly just use `hasValidExtension()`?

cmd/export.go
```diff
@@ -58,18 +88,18 @@ var exportCmd = &cobra.Command{
...
 		}
 		exportType, err := cmd.Flags().GetString("type")
-		exportType = strings.ToLower(exportType)
 		if err != nil {
 			return err
 		}
+		exportType = strings.ToLower(exportType)
+		if validExtension := providedExportTypes[exportType]; !validExtension {
+			return fmt.Errorf("export type not supported")
+		}
```
Gituser143, line 97: Here you're using the returned key to check if `exportType` exists. Wouldn't it be safer to use the second argument returned? Here's an [example](https://stackoverflow.com/questions/2050391/how-to-check-if-a-map-contains-a-key-in-go)

Gituser143 (changes requested):

cmd/export.go
MadhavJivrajani: Oh right, that's what I meant to do, my bad

cmd/export.go
MadhavJivrajani: > @MadhavJivrajani These two functions have the same signature, why not directly just use hasValidExtension()?


The two functions do have the same signature, but the reason I've separated them is so that additional file validation logic can be added without modifying the logic for checking extension types. Doesn't really make much of a difference right now, just separation of concerns.

Gituser143: @MadhavJivrajani do you have any other validation checks you plan to include here? If not, either directly use the `hasValidExtension()` func or maybe put a comment explaining TODO?

MadhavJivrajani: There's no other validation to be done right now and nothing TODO as well. Might as well keep them separate from now itself rather than refactoring later on.

Gituser143 (approved):

MadhavJivrajani:
@Gituser143 I just realised, for checking if export type is supported, you suggested changed it to `_, ok` but won't it be better to check the first value as done initially since that can ensure that export types that might be explicitly disabled through a config file are also taken care of.

Gituser143:
@MadhavJivrajani using the` _, ok` method would return a boolean value if present or not. You could then check the value if explicitly disabled. Then just use 
```go
val, ok := map[key]; !ok {
  // some error
} else if ok {
  //explicit cases with value
}
```

MadhavJivrajani:
Yeah, I know. That's not what I meant. 

Ideally, that would be the way to check, but since here the type of the values in the map are explicitly boolean, we can avoid two extra branches in the code by just checking the value returned. Also what exactly do you mean by "safer" in this context? I didn't quite get you there.

Gituser143:
> Also what exactly do you mean by "safer" in this context? I didn't quite get you there.

In the case where element exists, the boolean value is returned. But if it doesn't exist, If I'm right, `nil` is returned. So it would be safer to first act solely on whether or not the value exists by evaluation the boolean value right?

MadhavJivrajani:
If a key doesn't exist, the default value for the value type (int this case `false` for `bool`) will be returned and not `nil`

Gituser143:
It seems to return empty string, not false. You can verify [here](https://play.golang.org/p/uGgDUzXOsWv).

MadhavJivrajani:
The value type in that example is `string` and not `bool` xD

Gituser143:
Yea, it's the 0 value. You can directly work on value. It'll be fine. Sorry for the long deviation :P

Gituser143 (approved):

Issue #64, opened 2020-10-03
[REFACTOR] Small changes needed to increase code readability and maintainability.


MadhavJivrajani:
Small amounts of refactoring can be done to all files under `cmd` which have additional flags.

## Creating `const` types for default values of flags.
- Each flag in [cobra]() needs to have a default value set. As of now the default value in some cases is mentioned while creating the flag such as https://github.com/pesos/grofer/blob/03a98094cd2fd6b8822d0295f0e38e230e5ee730/cmd/proc.go#L106

- For increased readability and also maintainability, it would be better to have a `const` of these default values like [this](https://github.com/pesos/grofer/blob/03a98094cd2fd6b8822d0295f0e38e230e5ee730/cmd/proc.go#L31):
```go
const (
	DefaultProcRefreshRate = 3000
)
```

## Splitting flag initialization to multiple lines

A flag initialization such as:
```go
procCmd.Flags().Int32P("refresh", "r", DefaultProcRefreshRate, "Process information UI refreshes rate in milliseconds greater than 1000")
```
could be re-written to:
```go
procCmd.Flags().Int32P(
    "refresh",
    "r",
    DefaultProcRefreshRate,
    "Process information UI refreshes rate in milliseconds greater than 1000",
)
```

souvikmaji:
I am picking this up.

MadhavJivrajani:
Sure.

PR #67, opened 2020-10-04
refactored flag variable default values into consts

souvikmaji:
refers to  #64

### DIFF
--- a/cmd/proc.go
+++ b/cmd/proc.go
@@ -29,7 +29,8 @@ import (
 )
 
 const (
-	DefaultProcRefreshRate = 3000
+	defaultProcRefreshRate = 3000
+	defaultProcPid         = -1
 )
 
 // procCmd represents the proc command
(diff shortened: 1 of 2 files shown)


cmd/proc.go
```diff
@@ -102,6 +103,16 @@ func init() {
...
+		"r",
+		defaultProcRefreshRate,
+		"Process information UI refreshes rate in milliseconds greater than 1000",
+	)
+
+	procCmd.Flags().Int32P(
+		"pid",
+		"p",
+		defaultProcPid,
+		"specify pid of process")
```
MadhavJivrajani, line 117:
```suggestion
		"specify pid of process",
	)
```

MadhavJivrajani (changes requested):
- The check for valid PID can be done using the newly defined consts
```go
if pid != defaultProcPid {
    ...
}
```

- Feel free to remove the extra comments in the `init()` function.

souvikmaji:
What validation should i add for pids?

MadhavJivrajani:
Validation is already taken care of, since you've added the const `defaultProcPid`, you could use that in the check on [line 64](https://github.com/souvikmaji/grofer/blob/94f1682908b6d9571f44fe05dc307ee4ff94e818/cmd/proc.go#L64) so that if a default value should be changed, it only needs to be changed in the defined consts and not anywhere else.

MadhavJivrajani (approved):
Looks good! Thanks for your contribution! I'll wait for another review from either @Gituser143 or @metonymic-smokey before merging.

metonymic-smokey (approved):

MadhavJivrajani:
@souvikmaji a PR for changing PID types has already been made and ideally should not be a part of this PR. Please revert those changes for now.

souvikmaji:
Reverted

MadhavJivrajani (approved):

PR #71, opened 2020-10-04
Add per proc export functionality. [WIP]

Gituser143:
# Description

Implemented per process JSON export functionality. This builds on #63 
Currently the JSON output is stored fully in memory, waiting on @MadhavJivrajani for help :P

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# Checklist:

- [x] I have read the [contribution guidelines](https://github.com/pesos/grofer/blob/master/CONTRIBUTING.md) and followed it as far as possible. 
- [x] I have performed a self-review of my own code (if applicable)
- [x] I have commented my code, particularly in hard-to-understand areas (if applicable)
- [x] I have run `go fmt` on my code ([reference](https://blog.golang.org/gofmt))
- [x] I have made corresponding changes to the documentation (if applicable)
- [x] My changes generate no new warnings 
- [x] Any dependent and pending changes have been merged and published

PR #72, opened 2020-10-05
Memory buffering to reduce disk IO and optimize intermediate memory usage

MadhavJivrajani:
# Description
- Instead of storing the entire export profile array in memory, added "periodic" appends to export file.
- Periodic meaning the output of a certain number of iterations will be buffered in memory before being written to file.
- Export is not done every iteration as that would increase I/O overhead for large values of `iter`. 
- Buffering and error handling is done using buffered channels.
- The fraction of the total output to be buffered in memory is also added as a flag to the `export` command. 

## Type of change

# Checklist:

### DIFF
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -24,10 +24,11 @@ import (
 )
 
 const (
-	defaultExportRefreshRate = 1000
-	defaultExportIterations  = 10
-	defaultExportFileName    = "grofer_profile"
-	defaultExportType        = "json"
+	defaultExportRefreshRate    = 1000
+	defaultExportIterations     = 10
+	defaultExportBufferFraction = 0.5
+	defaultExportFileName       = "grofer_profile"
+	defaultExportType           = "json"
 )
 
 // Maintain a map of extensions provided by grofer.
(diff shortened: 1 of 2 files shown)


MadhavJivrajani:
@Gituser143, apologies for the delay, I will be AFK for the day now, in the meantime feel free to spam the PR with suggestions, comments, questions, or any form of feedback.

Gituser143:
@MadhavJivrajani this took me a while to get. Hopefully I understand it now. I have one small issue with it though, the output JSON is no longer an array of JSON objects, rather it's just JSON object after JSON object printed in sequence. If you tried verifying the JSON structure, it shows that it's invalid. You can verify the JSON integrity [here](https://jsonlint.com/).

Ideally we would want to have these enclosed in an array right?

Gituser143:
One solution I can think of is, use a `sync.Once` variable to open `[` before the first object write and a comma after every object. In the case where we receive a true `status.finished` value, we can close the array with `]`.  What do you think?

MadhavJivrajani:
> @MadhavJivrajani this took me a while to get. Hopefully I understand it now. I have one small issue with it though, the output JSON is no longer an array of JSON objects, rather it's just JSON object after JSON object printed in sequence. If you tried verifying the JSON structure, it shows that it's invalid. You can verify the JSON integrity here.

Ideally we would want to have these enclosed in an array right?

Ah damn, how did I miss this xD
Yes, you're right

MadhavJivrajani:
> One solution I can think of is, use a sync.Once variable to open [ before the first object write and a comma after every object. In the case where we receive a true status.finished value, we can close the array with ]. What do you think?

Yes, this is a good idea and most will work, I'll make the changes, thanks!

Gituser143:
> Yes, this is a good idea and most will work, I'll make the changes, thanks!
@MadhavJivrajani One thing you might have to keep in mind is how to put the comma. You can't put it after every object as it must be avoided for the last object. You could try putting it before we receive an object from the channel and avoid it for the very first time using the `sync.Once` variable. 

Or you might have an entirely different approach. try it out and let me know which works best, I'll update the per proc export features the same way by then!

MadhavJivrajani:
@Gituser143 I've made the changes, although I'm not all that satisfied with them, JSON validates fine, but still feels kinda hacky. I have another approach in mind which I will try and implement soon.

Gituser143:
@MadhavJivrajani I'm closing this along with #71. Will open a fresh PR

PR #74, opened 2020-10-08
Feature/help

### LINKED ISSUE #42: [FEATURE REQ] add functionality for getting help for a particular command

anihm136:
# Description

Adds a modular help widget, which can be used to add a help screen to any page with appropriate help string. Also adds help widget to existing pages

Fixes #42 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] This change requires a documentation update

# Checklist:

### DIFF
--- a/src/display/general/overallGraphs.go
+++ b/src/display/general/overallGraphs.go
@@ -129,8 +129,13 @@ func RenderCharts(ctx context.Context,
 				case "<Escape>":
 					helpVisible = false
 					updateUI()
+				case "j", "<Down>":
+					help.List.ScrollDown()
+					ui.Render(help)
+				case "k", "<Up>":
+					help.List.ScrollUp()
+					ui.Render(help)
 				}
-				ui.Render(help)
 			} else {
 				switch e.ID {
 				case "?":
@@ -303,8 +308,13 @@ func RenderCPUinfo(ctx context.Context,
 				case "<Escape>":
 					helpVisible = false
 					updateUI()
+				case "j", "<Down>":
+					help.List.ScrollDown()
+					ui.Render(help)
+				case "k", "<Up>":
+					help.List.ScrollUp()
+					ui.Render(help)
 				}
-				ui.Render(help)
 			} else {
 				switch e.ID {
 				case "?":
--- a/src/display/general/overallGraphs.go
+++ b/src/display/general/overallGraphs.go
@@ -90,7 +90,16 @@ func RenderCharts(ctx context.Context,
 		// Adjust Grid dimensions
 		myPage.Grid.SetRect(w/2, 0, w, h-heightOffset)
 
-		help.Resize(w,h)
+		help.Resize(w, h)
+
+		if helpVisible {
+			ui.Render(help)
+		} else {
+			ui.Render(myPage.Grid)
+			for i := 0; i < numCores; i++ {
+				ui.Render(myPage.CPUCharts[i])
+			}
+		}
 	}
 
 	updateUI() // Initialize empty UI
@@ -116,25 +125,16 @@ func RenderCharts(ctx context.Context,
 			if helpVisible {
 				switch e.ID {
 				case "?":
-					ui.Clear()
-					ui.Render(help)
+					updateUI()
 				case "<Escape>":
 					helpVisible = false
-					ui.Clear()
-					ui.Render(myPage.Grid)
-					for i := 0; i < numCores; i++ {
-						ui.Render(myPage.CPUCharts[i])
-					}
+					updateUI()
 				}
 				ui.Render(help)
 			} else {
 				switch e.ID {
 				case "?":
-					ui.Clear()
-					ui.Render(myPage.Grid)
-					for i := 0; i < numCores; i++ {
-						ui.Render(myPage.CPUCharts[i])
-					}
+					updateUI()
 				case "s": //s to pause
 					pause()
 				}
@@ -270,6 +270,11 @@ func RenderCPUinfo(ctx context.Context,
 		ui.Clear()
 		myPage.Grid.SetRect(0, 0, w, h)
 		help.Resize(w, h)
+		if helpVisible {
+			ui.Render(help)
+		} else {
+			ui.Render(myPage.Grid)
+		}
 	}
 
 	updateUI()
@@ -294,19 +299,16 @@ func RenderCPUinfo(ctx context.Context,
 			if helpVisible {
 				switch e.ID {
 				case "?":
-					ui.Clear()
-					ui.Render(help)
+					updateUI()
 				case "<Escape>":
 					helpVisible = false
-					ui.Clear()
-					ui.Render(myPage.Grid)
+					updateUI()
 				}
 				ui.Render(help)
 			} else {
 				switch e.ID {
 				case "?":
-					ui.Clear()
-					ui.Render(myPage.Grid)
+					updateUI()
 				case "s": //s to pause
 					pause()
 				}
--- a/src/display/general/overallGraphs.go
+++ b/src/display/general/overallGraphs.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	ui "github.com/gizak/termui/v3"
+	h "github.com/pesos/grofer/src/display/misc"
 	info "github.com/pesos/grofer/src/general"
 	"github.com/pesos/grofer/src/utils"
 )
@@ -32,6 +33,7 @@ import (
 var isCPUSet = false
 
 var run = true
+var helpVisible = false
 
 // RenderCharts handles plotting graphs and charts for system stats in general.
 func RenderCharts(ctx context.Context,
@@ -46,6 +48,8 @@ func RenderCharts(ctx context.Context,
 	var on sync.Once
 	var totalBytesRecv float64
 	var totalBytesSent float64
+	var help *h.HelpMenu = h.NewHelpMenu()
+	h.SelectHelpMenu("main")
 
 	// Get number of cores in machine
 	numCores := runtime.NumCPU()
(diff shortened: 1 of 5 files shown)


src/display/general/overallGraphs.go
```diff
@@ -80,14 +84,13 @@ func RenderCharts(ctx context.Context,
 		if isCPUSet {
 			for i := 0; i < numCores; i++ {
 				myPage.CPUCharts[i].SetRect(0, i*height, w/2, (i+1)*height)
-				ui.Render(myPage.CPUCharts[i])
```
Gituser143, line 83: @anihm136 Why was this removed?

src/display/general/overallGraphs.go
```diff
@@ -80,14 +84,13 @@ func RenderCharts(ctx context.Context,
...
 			}
 		}
 
 		// Adjust Grid dimensions
 		myPage.Grid.SetRect(w/2, 0, w, h-heightOffset)
 
-		ui.Render(myPage.Grid)
```
Gituser143, line 90: This too. These are required during resizes and updates.

Gituser143 (changes requested):
@anihm136 the deletions to `updateUI()`

A full screen render ends up looking like this. It only resizes during the tick.

Screenshot from 2020-10-08 21-45-53
 break the existing resizing and rendering. 

When switching back, it ends up looking like so (that is not a cropped image, that is the entire terminal :P):
Screenshot from 2020-10-08 21-48-16

We need the `ui.Render(mypage.grid)` back in the `updateUI()` function and the calls to `updateUI()` are necessary too.

EDIT: To recreate the issue, run `grofer -r 3000` and resize the terminal.

anihm136:
Okay looks like the issue occurs in between ticks right. I did not test that very well. The reason I had to remove that is coz it re-renders the main page over the help page on resize. I could have moved the same logic I'm using on keypress there also, but it seemed to be working fine without (since I did not test at the slower refresh rates). I'll move the logic into the update function also

Gituser143:
There might be a simpler fix for the re rendering of UI. You could call the pause function to stop the main UI rendering. And once the help dialog is closed, call the pause function again to resume rendering. Hopefully this might make your tasks easier to solve.

anihm136:
@Gituser143 should be sorted now, can you check?

Gituser143:
@anihm136 3 things you missed out:

* Use ui.clear() before rendering grid or help alone (when using `ui.Render()` by itself instead of `updateUI()`), else it causes a noticeable lag.
* You'll have to implement a scroll option for the help widget, as it clips off a few commands and there's no way to see that.
* Call the pause function before rendering the help widget as it ends up updating data unnecessarily (since it isn't visible) in the background.

Additionally, maybe use `q` along with `ESC` to exit the help widget and then exit the UI. This I'm not sure if it's really necessary, just felt it would be consistent. Maybe take up another opinion (or weigh in yourself) if you want to implement it.

src/display/general/overallGraphs.go
```diff
@@ -206,7 +235,12 @@ func RenderCharts(ctx context.Context,
 			}
 
 		case <-tick: // Update page with new values
-			updateUI()
+			if !helpVisible {
+				ui.Render(myPage.Grid)
+				for i := 0; i < numCores; i++ {
+					ui.Render(myPage.CPUCharts[i])
+				}
+			}
```
Gituser143, line 243: This could be shortened to a call to `updateUI()` itself right?

src/display/general/overallGraphs.go
```diff
@@ -253,8 +293,25 @@ func RenderCPUinfo(ctx context.Context,
 			case "<Resize>":
 				updateUI()
 
-			case "s": // s to stop
-				pause()
```
Gituser143, line 257: Why was this case removed?

src/display/process/allProcs.go
```diff
@@ -146,37 +155,54 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
...
-				previousKey = ""
-			} else {
-				previousKey = e.ID
+				ui.Render(myPage.Grid)
+				if previousKey == "g" {
+					previousKey = ""
+				} else {
+					previousKey = e.ID
+				}
 			}
```
Gituser143, line 206: This entire switch is a little messy, it's what's causing the non scrolling issue for the help widget. Maybe boil this down to An if condition and two switch cases? 

```go
if helpVisible {
  // handle keyboard input for the help bar.
  // switch case for e.ID
} else {
  // handle keyboard input for main menu.
  // switch case for e.ID
}
```
EDIT: This can be done for all the other pages too I feel.

src/display/general/overallGraphs.go
anihm136: It could, but it would cause unneccesary re-renders when the help menu is shown (since that does not need to be updated on every tick)

src/display/process/allProcs.go
anihm136: This is pretty much what's happening right now. The switch is inside the if condition to handle each individual key

src/display/general/overallGraphs.go
anihm136: It's not removed. I just moved it to be handled only when help menu is not being shown i.e, on the main page only

src/display/general/overallGraphs.go
anihm136: I can change it to updateUI there if you prefer, should not cause any issues

anihm136:
@Gituser143 First and second points should be fixed now. For the third point, as I mentioned in one of the comments, it does not update in the background because there's a condition checking if the help menu is visible and avoiding re-renders then
As for using `q` to leave the help menu, I personally feel it's more consistent for it to quit the application. However, I'm open to more opinions and if that feels more natural to a majority of users then it can be changed

src/display/general/overallGraphs.go
Gituser143: Yes, this was a mistake from my end. My bad!

src/display/process/allProcs.go
Gituser143: @anihm136 right now there are 3 different switches. They go like: 
```go 
case e := ui.Events:
  switch e.ID {
    // This handles close and resize
  }
  if helpVisible {
    // cases for help widget
  } else {
    // cases for main UI
  }

Comment (Gituser143) [reply src/display/process/allProcs.go]: This can be condensed is all.

**Comment by Gituser143** [comment]
@anihm136 There's an unnecessary empty line at the end main and proc_pid keybindings. 
image

It shows up as so when scrolled to the bottom. This can be removed.

Apart from this everything else looks great! I'll approve right after the extra newline thing is fixed!

**Comment by anihm136** [comment]
@Gituser143 should be fixed now

**Comment by Gituser143** [comment]
@anihm136 for some reason, the `grofer proc -p PID` is now broken. Can you verify?

File: src/display/process/procGraphs.go [hunk 6]
```diff
@@ -96,9 +98,18 @@ func ProcVisuals(endChannel chan os.Signal,
...
+		ui.Clear()
+		if helpVisible {
+			ui.Render(help)
+		} else {
+			ui.Render(myPage.Grid)
+		}
+
 	}
 
+	updateUI()
```
Gituser143, line 111: This breaks the UI as the PID table rows in the grid is not initialized. Removing this should fix the issue.

Gituser143 (changes requested):

src/display/process/procGraphs.go
anihm136: Done. Is that required in the other two graphs though, if it can be removed here?

Gituser143: I think it is, as it helps render an empty UI at first thus reducing lags.

Gituser143 (changes requested):
These look good! @anihm136 There are 1 or 2 places where you've commented out code. Could you please remove them?

MadhavJivrajani (commented):
Hi,
First off, this is great work @anihm136!

I had a query which I was hoping you could answer:
- Could `PROC_KEYBINDS` be made an array of strings, since that's what the `list` widget of `termui` accepts, the reason I ask this is simply from an efficiency point of view, I noticed that you are using the [`Split()`](https://golang.org/src/strings/strings.go?s=7164:7198#L289) function in the `Resize()`. Every time the `Resize()` function is called, `Split()` will iterate over the string as well as allocate an extra array which is ultimately returned, although this is garbage collected, we could maybe try and improve that aspect, and the same logic applies to the `Count()` function as well (you could use `len(arrayName)`). I'd like your opinion on this and whether you think it's feasible to implement without compromising the readability/simplicity of code.

MadhavJivrajani (changes requested):
- Could you add info about the keybinding to the README? 
- Now that #54 is merged, maybe add that keybinding for `grofer proc` as well.

anihm136:
> Hi,
> First off, this is great work @anihm136!
> 
> I had a query which I was hoping you could answer:
> - Could `PROC_KEYBINDS` be made an array of strings, since that's what the `list` widget of `termui` accepts

I think the main reason I used the string itself is for readability of the formatting. In the keybindings for proc you can see that there is some indentation, which might be confusing if implemented as an array of strings. If that's not an issue I can change it to an array of strings

MadhavJivrajani:
> I think the main reason I used the string itself is for readability of the formatting. In the keybindings for proc you can see that there is some indentation, which might be confusing if implemented as an array of strings. If that's not an issue I can change it to an array of strings

Hmm, I understand, my recommendation would still be to convert to an array of strings since that would be pre-allocated by the compiler and would not grow/shrink at run time and any overheads of garbage collection wrt to that `Split()` slice (pointer to an underlying array) won't have to be worried about and additionally, we might save a few clock cycles when the state of the program is in a "static" state, i.e. on the help window even though background fetches for data continue.

anihm136:
Okay, that was some insane merge conflict. @Gituser143 the commented code should be removed now, let me know if I've missed someplace. @MadhavJivrajani I've changed it to use a slice of strings instead of a string, and the new keybinding has been added as well. Only documentation is remaining

MadhavJivrajani (approved):
LGTM!
Great PR :smile:

Issue #75, opened 2020-10-09
[BUG] High exit time


Samyak2:
**Describe the bug**

`grofer` or `grofer proc` takes a few seconds to exit to the shell prompt after pressing `q` or `ctrl-c`.

The time taken to exit increases as the refresh time is increased.

**To Reproduce**

1. Run `./grofer -r 10000` in a terminal
2. Exit grofer using `q` or `ctrl-c`
3. Notice time taken for the shell prompt to reappear

**Expected behavior**

The process should exit immediately after `q` or `ctrl-c` is pressed.

**Screenshots**

![image](https://user-images.githubusercontent.com/34161949/95593263-15e26580-0a67-11eb-920d-0685b3c225b3.png)

**Desktop (please complete the following information):**
 - OS: Fedora 32 KDE
 - Go Version: v1.14.9

**Additional context**

[This comment in #54](https://github.com/pesos/grofer/pull/54/#issuecomment-706170043)

MadhavJivrajani:
On investigating a little, the problem seems to occuse because of this line:
https://github.com/pesos/grofer/blob/5e9d902f56595d407529d3bde056c7a636613f9b/src/general/generalStats.go#L67

ataboo:
May I take a look at this one?

MadhavJivrajani:
Absolutely! Have at it :smile:

Gituser143:
@ataboo Thanks for the fix!

PR #77, opened 2020-10-10
GH-75: using after channel instead of delay so exit is immediate

ataboo:
# Description

Using `time.After` channel instead of delay so that exit is immediate when application stops.

Fixes # GH-75

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

# Checklist:

### DIFF
--- a/src/display/process/procGraphs.go
+++ b/src/display/process/procGraphs.go
@@ -203,6 +203,7 @@ func ProcVisuals(ctx context.Context,
 					[]string{"[Nice value](fg:yellow)", strconv.Itoa(int(data.Nice))},
 					[]string{"[Thread count](fg:yellow)", strconv.Itoa(int(data.NumThreads))},
 					[]string{"[Child process count](fg:yellow)", strconv.Itoa(len(data.Children))},
+					[]string{"[Last Update](fg:yellow)", time.Now().Format("15:04:05")},
 				}
 				myPage.PIDTable.Title = " PID: " + strconv.Itoa(int(data.Proc.Pid)) + " "
 
--- a/cmd/proc.go
+++ b/cmd/proc.go
@@ -72,7 +72,7 @@ Syntax:
 			}
 
 			eg.Go(func() error {
-				return process.Serve(proc, dataChannel, ctx, int32(4*procRefreshRate/5))
+				return process.Serve(proc, dataChannel, ctx, int64(4*procRefreshRate/5))
 			})
 			eg.Go(func() error {
 				return procGraph.ProcVisuals(ctx, dataChannel, procRefreshRate)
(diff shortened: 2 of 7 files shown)


Gituser143:
@ataboo Thanks for the PR! This is a beautiful fix! But it needs to be implemented for other parts of the UI too. Take a look at other functions that serve data to the UI. A search for occurrences of `time.Sleep()` should point you at all the other functions.

ataboo:
Okay, I'll have a look around for other sleep calls.

Gituser143 (approved):
This looks great! Unless you plan to make any more changes, I'll wait for a second review before merging this PR.

ataboo:
Yeah I think that should cover this one.  Sounds good.

MadhavJivrajani (commented):
Oh wow, that was really interesting, I learned something new today :p 

I had a couple of questions which I was hoping you could answer:
- Could the 'after' channel be created before the for loop and then used in the `select`? The reason I ask is since doing `time.After(d)` creates a new channel every iteration which will have to be subsequently cleaned up.
- Could `time.NewTimer()` be used instead of `time.After()` due to the reason mentioned [here](https://golang.org/pkg/time/#After)?

src/display/process/procGraphs.go
```diff
@@ -203,6 +203,7 @@ func ProcVisuals(ctx context.Context,
 					[]string{"[Nice value](fg:yellow)", strconv.Itoa(int(data.Nice))},
 					[]string{"[Thread count](fg:yellow)", strconv.Itoa(int(data.NumThreads))},
 					[]string{"[Child process count](fg:yellow)", strconv.Itoa(len(data.Children))},
+					[]string{"[Last Update](fg:yellow)", time.Now().Format("15:04:05")},
```
ataboo, line 206: This was more for me to spot if things were actually updating. It can come out if not needed.

ataboo:
I think a ticker would fit the bill better if we're trying to prevent the performance hit for recreating the timer with After.

I consolidated the ticker with a callback for the action.  Note that I left a "last update" field in the individual proc display.  I think it'd be good for feedback and it was helpful for testing, but I can take it back out if you want.

MadhavJivrajani:
> Note that I left a "last update" field in the individual proc display. I think it'd be good for feedback and it was helpful for testing, but I can take it back out if you want.

It's probably not needed in the display.

> I think a ticker would fit the bill better if we're trying to prevent the performance hit for recreating the timer with After.
I consolidated the ticker with a callback for the action.

Yes, that is a better fit for our issue. Thank you! I learnt quite a lot from your PR :smile:

MadhavJivrajani (approved):
:100:

MadhavJivrajani:
Sorry about the revert commit reference, that was unintentional.

PR #84, opened 2020-12-09
Add JSON export functionality

Gituser143:
# Description

Added export functionality to export monitor/profile in line JSON format.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

# Checklist:


Gituser143:
@MadhavJivrajani, that was a fairly lorge merge, go through when at leisure :P

MadhavJivrajani (commented):
Any reason why errors aren't being handled in `generalExport.go`? 

`os.Remove` returns an error too, that should be handled too imo. Let me know if any specific reason for not doing so.

MadhavJivrajani:
Also, what happened to truncating the floating-point values in the export files to 2 decimal places?

Gituser143:
> Also, what happened to truncating the floating-point values in the export files to 2 decimal places?

This I've implemented only for disk and memory values, I wasn't able to for CPU as we directly fetch a slice of floats and the existing rounding function wouldn't work on that. I'll make a uniform one for all the metrics once we have a solid export approach

Gituser143:
> `os.Remove` returns an error too, that should be handled too imo. Let me know if any specific reason for not doing so.

I missed this, my bad. I'll fix it by the next commit. Any other places I've missed error handling?

MadhavJivrajani:
> I missed this, my bad. I'll fix it by the next commit. Any other places I've missed error handling?

Nope, just that and the corresponding place in `procExport.go` as well. Rest LGTM

MadhavJivrajani:
Wrt the approach for export, this looks good to me for the most part, the buffering approach might prove helpful, but from initial profiling, I wasn't able to find significant differences in the two approaches, at least this one keeps the code simple. 

Here's the profiling I had run:

`export_test.go`

```go
package general_test

import (
	"testing"

	g "github.com/pesos/grofer/src/export/general"
)

func TestExport(t *testing.T) {
	err := g.ExportJSON("test", 100, 10)
	if err != nil {
		return
	}
}
```

Command to run profiling:
```
go test -cpuprofile cpu.prof -bench . # replace cpuprofile with memprofile to profile for memory
```
and then 
```
go tool pprof file.prof
```
The `top n` (`n` being the top `n` most expensive nodes of the call graph to display info for) command or the `web` command might be helpful!

Maybe we can play around with the two approaches a little bit to get a better idea about which might be better. The buffered approach can be found [here](https://github.com/MadhavJivrajani/grofer/blob/b3074e26229e4a05b8e5f6bd7d04c7583817f0ee/src/export/general/export.go#L187).

Gituser143:
@MadhavJivrajani I've made appropriate changes and added a prompt to confirm profile overwriting. I might have to fix a few rounding off issues in per proc export, I'll do that if what's been done here is fine.

src/export/general/generalExport.go
```diff
@@ -174,7 +175,19 @@ func (data *OverallStats) updateData() error {
 // ExportJSON exports data to a JSON file for a specified number of iterations
 // and a specified refreshed rate.
 func ExportJSON(filename string, iter uint32, refreshRate uint64) error {
-	os.Remove(filename)
+	fmt.Println("Here", filename)
```
MadhavJivrajani, line 178: xDD

src/export/general/generalExport.go
```diff
@@ -174,7 +175,19 @@ func (data *OverallStats) updateData() error {
...
+	if _, err := os.Stat(filename); err == nil {
+		fmt.Printf("Previous profile with name 'grofer_profile' exists. Overwrite? (Y/N) ")
```
MadhavJivrajani, line 180: The filename can be something that the used might have used earlier as well, need not be `grofer_profile` right?

MadhavJivrajani (changes requested):

src/export/general/generalExport.go
Gituser143: I fixed, sorry about that xD

src/export/general/generalExport.go
Gituser143: Ah yes, I'll fix that

Gituser143:
@MadhavJivrajani that's done. I've moved the round off functions to `utils` and replicated the same process for `procExport` too. I might have missed something, do let me know if you find something.

MadhavJivrajani (changes requested):
Line 188 in `generalExport.go` also needs handling

Gituser143:
> Line 188 in `generalExport.go` also needs handling

This could be a temporary error, other iterations following this could work fine right? So we should probably not return entirely but maybe skip the iteration? Unless you feel aborting midway is better than skipped iterations...

MadhavJivrajani:
> > Line 188 in `generalExport.go` also needs handling
> 
> This could be a temporary error, other iterations following this could work fine right? So we should probably not return entirely but maybe skip the iteration? Unless you feel aborting midway is better than skipped iterations...

They could work fine, but it's highly probable that the reason that caused the encoding error might be due a file related issue rather than an encoding issue since we're passing an io writer to it. IMO it'd be better to terminate or provide some form of indication in the final output about the number of errors that occurred otherwise it might just be an empty file that got generated without an indication about what might have gone wrong.

Gituser143:
So maybe print non empty errors as encountered? This way we won't have to return entirely and if errors do pop up the user will be informed. Maybe not the best approach, but it's a start i guess?

MadhavJivrajani:
> So maybe print non empty errors as encountered? This way we won't have to return entirely and if errors do pop up the user will be informed. Maybe not the best approach, but it's a start i guess?

Yep, logging should work

Gituser143:
@MadhavJivrajani do we log the error instead of returning at `generalExport.go:184` too?

MadhavJivrajani:
Yeah, seems apt to do so 😄

Gituser143:
That seems about it, am I missing anything? @MadhavJivrajani

src/export/general/generalExport.go
```diff
@@ -177,15 +177,18 @@ func ExportJSON(filename string, iter uint32, refreshRate uint64) error {
...
 	encoder := json.NewEncoder(logFile)
 	stats := NewOverallStats()
-	var i uint32
 
-	for i = 0; i < iter; i++ {
+	for i := uint32(0); i < iter; i++ {
 		err := stats.updateData()
 		if err != nil {
-			return err
+			fmt.Println(err)
```
MadhavJivrajani, line 184: Shouldn't there be a `continue` here?

src/export/general/generalExport.go
```diff
@@ -177,15 +177,18 @@ func ExportJSON(filename string, iter uint32, refreshRate uint64) error {
...
 		}
 
 		err = encoder.Encode(&stats)
+		if err != nil {
+			fmt.Println(err)
```
MadhavJivrajani, line 189: `continue` here as well?

MadhavJivrajani (changes requested):
It might be a good idea to provide a little more info while printing the errors such as:
`error occurred in iteration x: <error>`

Gituser143:
> It might be a good idea to provide a little more info while printing the errors such as: error occurred in iteration x: <error>

This would happen by default right? With normal print it'll print as ErrorClass: Error Message?

src/export/general/generalExport.go
Gituser143: The only statements after are the sleep, it would be better to sleep after an error occurred than immediately attempt to refecth I felt. Even in the previous case, it would be better to sleep, but we can skip the second block if the first fails. Would that suffice?

Gituser143:
@MadhavJivrajani I've nested the if conditions and error handling, but I'm avoiding continue statements as that would skip the sleep. Additionally, I mentioned the iteration at which error occurs. The whole thing now looks like this:

```go
for i := uint32(0); i < iter; i++ {
		err := stats.updateData()
		if err != nil {
			fmt.Println("Error in iteration", i, "Error:", err)
		} else {
			err = encoder.Encode(&stats)
			if err != nil {
				fmt.Println("Error in iteration", i, "Error:", err)
			}
		}

		time.Sleep(time.Duration(refreshRate) * time.Millisecond)
	}
```

MadhavJivrajani (approved):
LGTM!

Gituser143:
@MadhavJivrajani I've made updates to README and commented code, merge if nothing's missing, else let me know what to change.

MadhavJivrajani (changes requested):
Since we aren't exporting as one big JSON, it might be helpful to mention in the README as well as the help section of the `export` command about the nature of export, i.e. we have one json object per line and not one big json object as this was done to prevent loading the entire json in memory while reading it as the resource monitoring data can get significantly large.

Gituser143:
That's fair, but how do you want it to be mentioned? This is specific to type of export right, I'll make a sub section for types?

MadhavJivrajani:
> That's fair, but how do you want it to be mentioned? This is specific to type of export right, I'll make a sub section for types?

Sounds good

Gituser143:
@MadhavJivrajani is this alright? Or do I make it under a different header entirely?

MadhavJivrajani (approved):
LGTM!

PR #102, opened 2021-04-29
Add Container metric functionality

### LINKED ISSUE #91: [FEATURE REQ] add monitoring of system resources given docker container id
Given that functionality for resource monitoring is provided on a per-process basis, I think it'd be super cool to have this feature for Docker containers as well. 

`grofer docker --cid cid`
`grofer export --cid cid`

**Proposed solution**
[Gopsutil](https://godoc.org/github.com/shirou/gopsutil/v3/docker) provides good support for getting info about containers. 

Beyond this, the extended `proc` functionality of gopsutil can be used by obtaining the PID from the container ID  by running something like `docker inspect -f '{{.State.Pid}}' <Container ID>`

**Additional context**
Need to see if information for all the currently provided widgets is obtainable.

Gituser143:
# Description

Add functionality to get overview of containers running by docker and related metrics. Can be accessed using `grofer container`.

Fixes #91 

## Type of change

# Checklist:


Gituser143:
@MadhavJivrajani I think I'm done here, pls review and let me know if I've missed/should change anything!

MadhavJivrajani (commented):
Screenshot_20210503_171846

I get this panic with `grofer container -c <cid of nginx>`
And the panic is not clean, command continue to go on in the background and is a little messy to kill and recover

Gituser143:
@MadhavJivrajani this is strange, I'm not able to reproduce this issue. Can you tell me what you ran?

MadhavJivrajani:
> @MadhavJivrajani this is strange, I'm not able to reproduce this issue. Can you tell me what you ran?

@Gituser143 it was:
```
docker run nginx
```
And then
```
grofer container -c <cid>
```

Gituser143:
@MadhavJivrajani I've added checks for instances of empty preCPU metrics. Let me know if this works. It should display NA in the Per CPU metrics as it won't be able to calculate usage without the previous CPU usage.

MadhavJivrajani:
> @MadhavJivrajani I've added checks for instances of empty preCPU metrics. Let me know if this works. It should display NA in the Per CPU metrics as it won't be able to calculate usage without the previous CPU usage.

The `panic` is fixed, although per cpu is always NA, not sure why yet. Will look more into this. 

There is another things, I tried mapping ports and for some reason it shows the port mapping twice:
image

Ans when I don't map any ports, it shows a mapping from host port `0` to container port `80`. The container part makes sense since nginx exposes that port, but port `0` doesn't make too much sense imo. 
image

Gituser143:
@MadhavJivrajani This is super weird, the port mappings seem to work fine for me, even the per CPU percent. Could this maybe be an issue with docker versions? 

image

Gituser143:
@MadhavJivrajani @metonymic-smokey I've made changes to handle docker engine not running errors and replaced uses of `ContainerStatsOneShot()` with `ContainerStats()`. The latter should fix the issues with per CPU utilisation. Let me know if there's anything else I've missed.

MadhavJivrajani:
@Gituser143 awesome, so CPU percentage seems to be updating well when I load the nginx container with requests. As per the double port mapping that are coming, I think I see the issue, so when I do `docker inspect <cid>`:
```json
"Ports": {
                "80/tcp": [
                    {
                        "HostIp": "0.0.0.0",
                        "HostPort": "8080"
                    },
                    {
                        "HostIp": "::",
                        "HostPort": "8080"
                    }
                ]
            }
```
The double appearence of `8080:80` seems to be appearing because of a seperation of IPv4 and IPv6. 
Docker version - 20.10.6
Docker API version - 1.41

Can you share what `docker inspect` gives you? 
Also, if this is an issue due to verions, we can probably restrict the mapping to IPv4? To check for that the Go `net` pacakge has a few utils iirc.

Gituser143:
@MadhavJivrajani Port mappings aren't derived from the `ContainerInspect()`, I get them directly from `ContainerList()` itself. The data returned, has ports within it as a field. The type can be found [here](https://pkg.go.dev/github.com/docker/docker@v20.10.6+incompatible/api/types#Container)

MadhavJivrajani:
Or instead of just IPv4, we could have another column in that table called `HostIP`, that would make things much clearer.

Gituser143:
> Or instead of just IPv4, we could have another column in that table called `HostIP`, that would make things much clearer.

This sounds like a good idea, clears more things up!

MadhavJivrajani:
> @MadhavJivrajani Port mappings aren't derived from the ContainerInspect(), I get them directly from ContainerList() itself. The data returned, has ports within it as a field. The type can be found here

Regardless, I would assume that the `inspect` command too gets its info from the API, just like `ContainerList` or `ContainerStats`

Gituser143:
> Regardless, I would assume that the `inspect` command too gets its info from the API, just like `ContainerList` or `ContainerStats`

They are different endpoints, thus have different format entirely.

MadhavJivrajani:
> They are different endpoints, thus have different format entirely.

Right, but whenever you would want to list a particular resource (like ports in this case), it would give a unifrom result, albeit in different format, is what I meant.

src/container/container.go
```diff
@@ -0,0 +1,272 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/client"
+	ui "github.com/gizak/termui/v3"
+	"github.com/pesos/grofer/src/general"
+)
+
```
MadhavJivrajani, line 31: Can you add a godoc style comment for exported structs?

src/container/container.go
```diff
@@ -0,0 +1,272 @@
...
+	// Get Container Stats
+	stats, _ := cli.ContainerStats(ctx, cid, false)
+	data := types.StatsJSON{}
+	err = json.NewDecoder(stats.Body).Decode(&data)
+	if err != nil {
+		return metrics, err
+	}
+
+	// Calculate CPU  and per CPU percent
```
MadhavJivrajani, line 133: Can calculation of CPU% be in a different function? We could write tests for it that way.

src/container/container.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	// Get processes in container
+	procs, err := cli.ContainerTop(ctx, cid, []string{})
+	if err != nil {
+		return metrics, nil
+	}
+
+	procData := []procInfo{}
+	for _, proc := range procs.Processes {
```
MadhavJivrajani, line 244: `proc[0]`, etc isn't very understandable. My suggestion would be to have different functions called `get<whatever proc[0] is supposed to give>()` or something similar. The reason I suggest seperate functions is mainly because the implementation might change in the future, in case it does, incorporating that change won't be too much of a hastle. 

But, if you think that having it the way it is is better, then it might be a good idea to maybe provide comments on what those fields exactly are.

src/container/overall.go
```diff
@@ -0,0 +1,149 @@
...
+		return
+	}
+	data := types.StatsJSON{}
+	err = json.NewDecoder(stats.Body).Decode(&data)
+	if err != nil {
+		return
+	}
+
+	// Calculate CPU percent
+	cpuPercent := 0.0
```
MadhavJivrajani, line 106: Same comment about CPU percentage as before.

src/utils/dataFormat.go
```diff
@@ -65,6 +65,19 @@ func RoundValues(num1, num2 float64) ([]float64, string) {
 		units = " per trillion "
 	}
 
+	if inBytes {
```
MadhavJivrajani, line 68: Can a few unit tests be written for this?

src/container/container.go
Gituser143: The docker engine API is defined this way, I see no reason why we'd want to create functions just to extract these values. As for the fields, they're named in the struct itself, so that should be clear in itself right?

src/utils/tickutils.go
```diff
@@ -35,7 +35,6 @@ func TickUntilDone(ctx context.Context, refreshRate int64, action func() error)
 		case <-ctx.Done():
 			return ctx.Err() // Stop execution if end signal received
 		case <-ticker.C:
-			break
```
MadhavJivrajani, line 38: I'm not sure I understand this, the `case` has no statements?

src/utils/dataFormat.go
Gituser143: Yea this can be done. Will do along with CPU percent calculations

src/utils/tickutils.go
Gituser143: The break is useless, as it exits from the select. Which will happen even without it. The ticker is just to exit the blocking select call and execute action.

Gituser143: leaving it generated a warning, hence the removal.

MadhavJivrajani: Gotcha. Can you add a comment for each of those cases? It took me a lil while to realise it.

MadhavJivrajani:
lgtm other than the above comments.

Gituser143:
@MadhavJivrajani, Okay so for the CPU metrics calculation, making it a function might not be the best idea because, it requires quite a lot of parameters and simply sending each of those individual metrics doesn't seem worth it. The parameters we'd have to send are:
- numCPUs
- CPUStats.CPUUsage
- PreCPUStats.CPUUsage
- CPUStats.SystemUsage
- PreCPUStats.SystemUsage

Each of which are a type defined externally. So I feel it would be better to leave this as is. What do you think?

MadhavJivrajani:
From what I can tell, all of them are available in the `StatsJSON` type. Sending a pointer to that struct to avoid copying sounds good enought to me

Gituser143:
> From what I can tell, all of them are available in the `StatsJSON` type. Sending a pointer to that struct to avoid copying sounds good enought to me

But if we're doing this, writing tests might get a little messy as we'd have to either fetch a similar stat or populate a struct ourself. Which when I come to think of it, shouldn't be much of an issue. I'll get to it :P

src/display/container/init.go
```diff
@@ -0,0 +1,307 @@
...
+	page.PortMapTable.ColResizer = func() {
+		x := page.PortMapTable.Inner.Dx()
+		page.PortMapTable.ColWidths = []int{
+			3 * x / 10,
+			2 * x / 10,
+			2 * x / 10,
+			2 * x / 10,
+		}
+	}
+	page.PortMapTable.Header = []string{"IP", "Host", "Container", "Type"}
```
MadhavJivrajani, line 265: Small nit, but can you change `IP` to `HostIP`? Just to be as clear as possible and consistent with the docker API.

MadhavJivrajani (changes requested):

src/container/container.go
Gituser143: Done!

src/container/container.go
Gituser143: Done and tests added!

src/container/overall.go
Gituser143: Changes made!

src/utils/tickutils.go
Gituser143: Yep, will do!

Gituser143:
@MadhavJivrajani changes have been made! Do let me know if I've missed/messed something up.

MadhavJivrajani (approved):
tenor

PR #106, opened 2021-05-06
Add Dockerfile

### LINKED ISSUE #101: [FEATURE REQ] Add Dockerfile for `grofer`
Have a Dockerfile for grofer for ease of use. 
```
docker run pesos/grofer:latest
```

**Additional context**
- On initial thought, this can be done by mounting the file systems being used by `gopsutil` to fetch metrics, `/proc` and `/sys` might be a good starting point.

Samyak2:
# Description

~~**It isn't working yet, do not merge pls**~~
It works now!

Fixes #101 

 - Added a Dockerfile based on the [`golang` image](https://hub.docker.com/_/golang)
 - ~~`grofer` is installed using `go get` (not from local source, I think?)~~ `grofer` is installed using the local current directory.
 - `HOST_PROC`, `HOST_SYS`, etc. is set for `gopsutil`
 - The whole root directory of the host has to be mounted as read only for this to work. A few other docker flags are also required.
 - See updated README for instructions

~~What does not work:~~

~~`grofer` just freezes on launch, while taking 100% CPU. It has to be stopped using `docker stop` (ctrl-c does not work). Only `--help`, `about` seem to work (`proc` freezes too). I tried setting `HOST_SYS`, etc. the issue still exists. I'm opening this draft PR to see if we can fix this.~~
 
## Type of change

# Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,16 +2,23 @@
 #             and Debian buster
 FROM golang:1-buster
 
-RUN go get -u github.com/pesos/grofer
+# ADD grofer /go/bin/
+ADD . /src
+
+WORKDIR /src
+
+RUN go install
+
+WORKDIR /
 
 # host's root (/) is assumed to be mounted at /host
 # set env vars for gopsutil to find the right files
 # refer: https://github.com/shirou/gopsutil#usage
 ENV HOST_PROC /host/proc
-# ENV HOST_SYS /host/sys
-# ENV HOST_ETC /host/etc
-# ENV HOST_VAR /host/var
-# ENV HOST_RUN /host/run
-# ENV HOST_DEV /host/dev
+ENV HOST_SYS /host/sys
+ENV HOST_ETC /host/etc
+ENV HOST_VAR /host/var
+ENV HOST_RUN /host/run
+ENV HOST_DEV /host/dev
 
 ENTRYPOINT ["grofer"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # we're using Golang 1.x (1.16.3 at the time of writing)
 #             and Debian buster
-FROM golang:1-buster
+FROM golang:alpine
 
 # ADD grofer /go/bin/
 ADD . /src
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,26 @@
 # we're using Golang 1.x (1.16.3 at the time of writing)
-#             and Debian buster
+#             and alpine
 FROM golang:alpine
 
+# - this section can be used to use the binary directly
+#   instead of building it inside the container.
+#   This binary will have to be built with
+#   `CGO_ENABLED=0 go build`
+# - you might need this if you're developing grofer (making
+#   changes) and want to test these in docker. Building
+#   grofer from source inside the container can be slow and
+#   leads to a wastage of time as well as (internet) bandwith
+#   due to `go get` being run everytime.
+# - uncomment the following line and comment out the next
+#   section to enable this
 # ADD grofer /go/bin/
+
+# - this section uses the current directory (grofer repo) and
+#   builds grofer inside the container which is then used when
+#   the container is run
 ADD . /src
-
 WORKDIR /src
-
 RUN go install
-
 WORKDIR /
 
 # host's root (/) is assumed to be mounted at /host
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+# we're using Golang 1.x (1.16.3 at the time of writing)
+#             and Debian buster
+FROM golang:1-buster
+
+RUN go get -u github.com/pesos/grofer
+
+# host's root (/) is assumed to be mounted at /host
+# set env vars for gopsutil to find the right files
+# refer: https://github.com/shirou/gopsutil#usage
+ENV HOST_PROC /host/proc
+# ENV HOST_SYS /host/sys
+# ENV HOST_ETC /host/etc
+# ENV HOST_VAR /host/var
+# ENV HOST_RUN /host/run
+# ENV HOST_DEV /host/dev
+
+ENTRYPOINT ["grofer"]
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -34,21 +34,29 @@ jobs:
       - name: Build image
         run: docker build . --file Dockerfile --tag $IMAGE_NAME
 
-      - name: Log into registry
-        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u ${{ github.actor }} --password-stdin
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Push image
         run: |
-          IMAGE_ID=docker.pkg.github.com/${{ github.repository }}/$IMAGE_NAME
-          # Change all uppercase to lowercase
-          IMAGE_ID=$(echo $IMAGE_ID | tr '[A-Z]' '[a-z]')
+          # repo owner - converted to lowercase
+          OWNER=$(echo "${{ github.repository_owner }}" | tr "[:upper:]" "[:lower:]")
+
           # Strip git ref prefix from version
           VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
           # Strip "v" prefix from tag name
           [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+          # if it's a PR, use `pr-<PR number>` as the tag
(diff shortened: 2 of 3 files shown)


Gituser143:
I just tried this out but used `golang` as image and my system just shut down xD

Samyak2:
> I just tried this out but used `golang` as image and my system just shut down xD

Oof, that's bad. Does buster work? 

> Edit: @Samyak2 any reason for picking buster?

No particular reason other than it being the latest one. I thought it would be good to lock it to a particular distro so that any dependencies (we don't have any yet) installed wouldn't break if the tag is updated.

Gituser143:
@Samyak2 Plis update the branch with `main` before continuing work as we had ignored a zero division error previously which caused quite a bit of trauma for me to catch :P

Samyak2:
So I tried the latest version, some things works now. `grofer` works well - CPU, RAM and disks are correct. Network does not look correct, `RX Total` shows `5.7 B` and the network graph does not show a spike when a speed test is running (`grofer` running on host does show this).

`grofer proc` shows only the container's processes i.e., only `grofer` is shown in the list. I confirmed that `HOST_PROC` was set in the container and also tried running it using `HOST_PROC=/host/proc` inside the container, the result is the same.

MadhavJivrajani:
@Samyak2 I believe the `grofer proc` issue can be fixed by running the container with the `--privileged` flag like `cAdvisor` does it. 
I haven't gotten around to investigating the network issue yet, will keep you posted.

MadhavJivrajani:
Also, @Samyak2, is it possible to use `golang:alpine` instead of `golang:1-buster`?
Sizes:

| Image | Size |
| -------- | ------ |
| `golang:alpine` | 651mb |
| `golang:1-buster` | 1.21gb |

Gituser143:
Also, I'm a little worried about how we'll perform kill operations. The signal can't be sent from within the container to kill the host process right? Whereas this might not be an issue with the containers, which reminds me, we've got to mount the docker socket too!

MadhavJivrajani:
> Also, I'm a little worried about how we'll perform kill operations. The signal can't be sent from within the container to kill the host process right?

While doing `docker run`, you can pass a flag saying `--pid=host` and it should work fine. 
> we've got to mount the docker socket too!

Oh good point

Gituser143:
> While doing docker run, you can pass a flag saying --pid=host and it should work fine.

This is so cool!

Samyak2:
> @Samyak2 I believe the `grofer proc` issue can be fixed by running the container with the `--privileged` flag like [`cAdvisor`](https://github.com/google/cadvisor#quick-start-running-cadvisor-in-a-docker-container) does it.

Ooh yes, that fixes it.

---

> Also, @Samyak2, is it possible to use `golang:alpine` instead of `golang:1-buster`?
> Sizes:
> Image 	Size
> `golang:alpine` 	651mb
> `golang:1-buster` 	1.21gb

@MadhavJivrajani The reason I did not choose alpine was because the [dockerhub page](https://hub.docker.com/_/golang) mentions

> This variant is highly experimental, and not officially supported by the Go project (see golang/go#19938 for details).

But if grofer works, it should provide a significant improvement.

---

> > Also, I'm a little worried about how we'll perform kill operations. The signal can't be sent from within the container to kill the host process right?
> 
> While doing `docker run`, you can pass a flag saying `--pid=host` and it should work fine.

This works well! Amazing

Samyak2:
Looks like alpine works, I'll switch to it.

MadhavJivrajani:
> Network does not look correct, RX Total shows 5.7 B and the network graph does not show a spike when a speed test is running (grofer running on host does show this).

@Samyak2 can you try this out again by passing `--network=host` flag while doing a `docker run` and see if that helps?

Samyak2:
> @Samyak2 can you try this out again by passing `--network=host` flag while doing a `docker run` and see if that helps?

@MadhavJivrajani It works! The network graph is consistent between host and docker :partying_face: 

---

I have a question though, what does `RX` and `TX Total` represent in the graph? It does not seem to change significantly over time, even when I run a speedtest that takes at least a few hundred megabytes (the graph does show a spike though).

Gituser143:
RX is received and TX is transmitted, so they represent the amount of data sent and received through network.

Dockerfile
```diff
@@ -0,0 +1,24 @@
+# we're using Golang 1.x (1.16.3 at the time of writing)
+#             and Debian buster
+FROM golang:alpine
```
MadhavJivrajani, line 3: I guess this comment needs to be updated after shifting to `golang:alpine`

Samyak2:
> RX is received and TX is transmitted, so they represent the amount of data sent and received through network.

Makes sense.

> Edit: I'm not 100% sure about this, but even with a speed test, at max a few mB gets used. It shouldn't really consume a lot of data.

I think it consumes at least a few hundred Mbs. Considering a speed of `100Mbps` and the test running for 10s, it would mean `100 * 2^20 * 10` bits or `125` MB consumed.

---

So I compared this with [`bpytop`](https://github.com/aristocratos/bpytop).

Before speed test, bpytop shows
```
Download Total: 12.5 GiB
```
and grofer shows
```
RX Total: 13.3 mB
```

---

After speed test, bpytop:
```
Download Total: 13.1 GiB
```
grofer:
```
RX Total: 13.9 mB
```

I think something is wrong here. I can open a new issue for this and move the discussion there.

Dockerfile
Samyak2: Oh right, good catch.

Samyak2:
Also, I had another idea related to this PR. We could publish this image to the [GitHub container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) and use actions to [automatically publish on release](https://docs.github.com/en/actions/guides/publishing-docker-images#publishing-images-to-github-packages).

I'll try making a workflow for this.

MadhavJivrajani:
That's a great idea

MadhavJivrajani:
@Samyak2 how's it going?

Samyak2:
Hi. I haven't been able to work on this over the past few days, I'll try to get the container workflow ready in a couple of days.

Samyak2:
I was able to write a workflow for building and publishing Docker images to the [GitHub Docker registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-docker-registry) (NOT the [container registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)) using [this starter workflow](https://github.com/actions/starter-workflows/blob/68d67073c3864b2836a59ed0af5b30d2ad3ff719/ci/docker-publish.yml) mentioned by @Gituser143 (thank you!).

Currently, it also publishes on PR. This was just to test the workflow. [act](https://github.com/nektos/act) failed at the registry login step, so I'm assuming `act` does not yet handle that. I'll remove that section once I confirm that the workflow works :)

Samyak2:
```
denied: Resource not accessible by integration
```

Wut. Does the docker registry need to be enabled somewhere before using it?

Samyak2:
In the latest commit, I changed the registry to the GitHub container registry since it's newer. This is based on [RocketChat's workflow](https://github.com/RocketChat/Rocket.Chat/blob/21fc380bee962e79b6957e209dc896677aded282/.github/workflows/build_and_test.yml#L280) for the same.

I'm assuming this will fail too. According to [this section](https://github.com/marketplace/actions/docker-login#github-container-registry) in `docker/login-action`'s documentation write access must be provided to this action. Do one of you @MadhavJivrajani @Gituser143 have access to this?

Gituser143:
> Do we go about enabling this or use the GitHub docker registry instead?

Yea I feel like we should. Maybe bring it up on the slack channel once?

Samyak2:
New error now:
```
denied: installation not allowed to Create organization package
```

Perhaps we need to provide write access to this Action as described [here](https://github.com/marketplace/actions/docker-login#github-container-registry).

[Relevant thread in the GitHub community](https://github.community/t/cant-authenticate-with-actions-token-for-pr-event/170752).

---

We could try doing this from an internal branch instead of a fork, as we won't need containers for every PR. I will direct this PR to a new, internal branch and we can test the Actions there. Is that ok?

MadhavJivrajani:
> We could try doing this from an internal branch instead of a fork, as we won't need containers for every PR. I will direct this PR to a new, internal branch and we can test the Actions there. Is that ok?

Sounds good 👍🏻 
Thanks!

Samyak2:
I have changed the base branch to `dockerfile`. I'll merge this now and open a PR from that branch?

MadhavJivrajani:
Thanks @Samyak2, just one thing before you proceed: you'll need to explicitly mount the docker socket for `grofer container` to work, that part of `-v /var/run/docker.sock:/var/run/docker.sock:ro` needs to be added in the README, if the command gets too long for the README, feel free to format it as well 

Post that feel free to merge it and open a PR from the new branch 😄

Samyak2:
Ah right, I forgot about that. I'll add the docker socket and split it into multiple lines.

Gituser143 (approved):

Gituser143:
Changes look good to me! We'll have to fix push soon though.

Samyak2:
Yes, I'll fix that from the internal `dockerfile` branch.

PR #109, opened 2021-05-08
Add Sort functionality to overall process and container tables

### LINKED ISSUE #107: [FEATURE REQ] Add keybindings for sorting output for `grofer proc`
As @Gituser143 pointed out, it'd be nice to have sort functionality based on different columns of the `grofer proc` output. 

`ctrl + 1` for sorting by first column of the output? And so on for the other columns. 

**Describe alternatives you've considered if any**
`shift` was discussed, but `termui` does not have keybindings for that.

Gituser143:
# Description

- Add sort capabilities to proc table and container table. 
- Data can be sorted on column numbers. Eg: 1 to sort on first column in ascending order and F1, to sort in descending.
- `grofer proc` can be sorted on all 8 fields
- `grofer container` can be sorted on all fields except Net & Block I/O (Multiple values exist, unsure on how to sort).

Fixes #107 

## Type of change

# Checklist:

### DIFF
--- a/src/utils/sort.go
+++ b/src/utils/sort.go
@@ -21,7 +21,10 @@ import (
 	"strconv"
 )
 
-// SortData helps sort table rows. It sorts the table based on values given in the sortIdx column and sorts ascending if sortAsc is true. sortCase is set to identify the set of 'less' functions to use to sort the selected column by.
+// SortData helps sort table rows. It sorts the table based on values given
+// in the sortIdx column and sorts ascending if sortAsc is true.
+// sortCase is set to identify the set of 'less' functions to use to
+// sort the selected column by.
 func SortData(data [][]string, sortIdx int, sortAsc bool, sortCase string) {
 
 	// Define less functions
@@ -57,24 +60,24 @@ func SortData(data [][]string, sortIdx int, sortAsc bool, sortCase string) {
 	switch sortCase {
 	case "PROCS":
 		sortFuncs = map[int]func(i, j int) bool{
-			0: intSort,
-			1: strSort,
-			2: floatSort,
-			3: floatSort,
-			4: strSort,
-			5: strSort,
-			6: strSort,
-			7: intSort,
+			0: intSort,   // PID
+			1: strSort,   // Command
+			3: floatSort, // CPU %
+			2: floatSort, // Memory %
+			4: strSort,   // Status
+			5: strSort,   // Foreground
+			6: strSort,   // Creation Time
+			7: intSort,   // Thread Count
 		}
 	case "CONTAINER":
 		sortFuncs = map[int]func(i, j int) bool{
-			0: strSort,
-			1: strSort,
-			2: strSort,
-			3: strSort,
-			4: strSort,
-			5: floatSort,
-			6: floatSort,
+			0: strSort,   // ID
+			1: strSort,   // Image
+			2: strSort,   // Name
+			3: strSort,   // Status
+			4: strSort,   // State
+			5: floatSort, // CPU %
+			6: floatSort, // Memory %
 		}
 
 	default:
--- a/src/utils/sort.go
+++ b/src/utils/sort.go
@@ -1,0 +1,86 @@
+/*
+Copyright © 2020 The PES Open Source Team pesos@pes.edu
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


MadhavJivrajani (changes requested):
If possible write some unit tests for the `SortData` functionality.

src/utils/sort.go
```diff
@@ -0,0 +1,86 @@
...
+*/
+
+package utils
+
+import (
+	"sort"
+	"strconv"
+)
+
+// SortData helps sort table rows. It sorts the table based on values given in the sortIdx column and sorts ascending if sortAsc is true. sortCase is set to identify the set of 'less' functions to use to sort the selected column by.
```
MadhavJivrajani, line 24: Small nit, but maybe break this doc comment into multiple lines?

src/utils/sort.go
```diff
@@ -0,0 +1,86 @@
...
+func SortData(data [][]string, sortIdx int, sortAsc bool, sortCase string) {
+
+	// Define less functions
+	intSort := func(i, j int) bool {
+		x, _ := strconv.Atoi(data[i][sortIdx])
```
MadhavJivrajani, line 29: Would ignoring the error be a good idea? Not sure of a case when such an error would occur, but if it does, ignoring it might not be the best thing. 

Since such a function assumed that its a valid `int` or a valid `float`, it might help to enforce this assumption by passing converted data to the sort funtion, or a wrapper that would potentially help handling ignored errors. WDYT?

src/utils/sort.go
```diff
@@ -0,0 +1,86 @@
...
+		if sortAsc {
+			return data[i][sortIdx] < data[j][sortIdx]
+		}
+		return data[i][sortIdx] > data[j][sortIdx]
+	}
+
+	floatSort := func(i, j int) bool {
+		x1 := data[i][sortIdx]
+		y1 := data[j][sortIdx]
+		x, _ := strconv.ParseFloat(x1[:len(x1)-1], 32)
```
MadhavJivrajani, line 47: Why the `len - 1`? Won't it mess up the conversion? 

Ref: https://play.golang.org/p/nr1JvhpmS-d

src/utils/sort.go
```diff
@@ -0,0 +1,86 @@
...
+		y, _ := strconv.ParseFloat(y1[:len(y1)-1], 32)
+		if sortAsc {
+			return x < y
+		}
+		return x > y
+	}
+
+	// Set function map
+	sortFuncs := make(map[int]func(i, j int) bool)
```
MadhavJivrajani, line 56: Might help mentioning where one can look to get the column names (in this case display/allProcs.go)

src/utils/sort.go
Gituser143: To remove the % symbol

src/utils/sort.go
Gituser143: But why? It's clean when reading the doc this way right?

MadhavJivrajani: Not really, you have to scroll and all, gets weird, maybe it's neat cause it wraps it in your editor?

src/utils/sort.go
Gituser143: I'm unsure how that helps here. Because the function is called in the same places where headers are defined, do you want to explicitly add a comment for both proc and container?

src/utils/sort.go
Gituser143: So newlines would fix the scroll? This is weird :P

MadhavJivrajani: Don't add new lines, break the doc comment into multiple lines to decrease length per line XD

src/utils/sort.go
Gituser143: You can't really give converted data to the `less` functions, you can only ensure data coming into the `sortData()` is clean. But even in the case that conversion fails, zero values are returned and the data will stay the same post sort. So I feel it's alright to ignore the error here. Handling it would cause unnecessary complexity to a simple function I feel.

src/utils/sort.go
Gituser143: Yea I got it xD

src/utils/sort.go
MadhavJivrajani: Fair enough 👍

src/utils/sort.go
Gituser143: In that case, I might as well mention which column is what field as a comment in the map pairs right. Do I do that or just link to headers directly?

MadhavJivrajani: Yeah that works too, basically provide some indication on what fields you're working with 👍

Gituser143:
@MadhavJivrajani I'm quite clueless on what to write for tests, because we're practically giving string slices of fixed formats to the function and it calls an internal function `sort.Slice()`. I'm not sure what to exactly test here. What do you think I should do?

MadhavJivrajani (approved):
LGTM ❤️

Issue #80, opened 2020-10-13
[FEATURE REQ] provide support for different signal types


MadhavJivrajani:
**Is your feature request related to a problem? Please describe.**
`gopsutil` provides [support](https://godoc.org/github.com/shirou/gopsutil/process#Process.SendSignal) for 4 different types of signals that can be sent to a process, signals other than these 4 need to be implemented such as `SIGABRT`. A list of signals can be obtained by running `kill -l`

**Describe the solution you'd like**
Maybe the implementation of sending signals could be done using the [`Signal()`](https://godoc.org/os#Process.Signal) function.

Samyak2:
Do these have to be implemented in separate functions (such as `sendHup`, `sendInt`) or a general function to send a given signal to a process?

MadhavJivrajani:
Separate functions can be implemented to send individual signals, but only a single exposed interface should preferably be available. We could have codes for each type of signals as consts and a `map` from these consts as keys and values as functions. Maybe we could get the function from the `map` and then call it, handle errors from it, etc.

anihm136:
I can try this. syscall already has aliases for all kinds of signals, which can be passed directly to Signal(). Do you think it's a good idea to forego gopsutil entirely for this functionality to maintain a uniform implementation?

MadhavJivrajani:
Yep! That would be ideal. Go for it!

anihm136:
Quick update, I've started work on this. `gopsutil` has a `SendSignal` method that takes an arbitrary signal as defined by `syscall` and sends it to a process. I'm planning to use it for further work on this

PR #112, opened 2021-05-14
Allow sending various signals to processes

### LINKED ISSUE #79: [FEATURE REQ] Implement a menu of signal type options for killing a process
Currently, the functionality for killing a process is based on sending a `SIGKILL` to the process. Instead of always sending a `SIGKILL`, it might be better to provide the user with a list of signal types that can be used to terminate the process, eg: `SIGTERM`, `SIGSTOP`, etc.

`gopsutil` provides [support](https://godoc.org/github.com/shirou/gopsutil/process#Process.SendSignal) for 4 different types of signals that can be sent to a process. Support for signals other than the available ones can also optionally be added. A new menu can be displayed when the keys for "killing" a process are pressed, this menu can display the signal options as a `termui list`. Based on the selected option, that particular signal should be sent. 

**Additional context**
Run `htop` and press `F9`  to kill a process, a menu of signal options is provided to choose from.

anihm136:
# Description

This PR aims to integrate sending of various signals to processes from the `grofer proc` subcommand. Currently, the command only allows killing of a process. The scope of this PR include -
- [x] Replacing Kill with SendSignal
- [x] Submenu to choose a signal to send to the selected process
- [x] Keybinding to launch submenu

Fixes #79, #80

## Type of change

# Checklist:


anihm136:
On second thoughts, I should have probably made this a draft PR. Just to clarify, this is still WIP and not ready for review yet. Will update when I'm done

src/display/process/allProcs.go
```diff
@@ -273,7 +274,7 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 						procToKill, err := proc.NewProcess(pidToKill)
 						myPage.ProcTable.CursorColor = selectedStyle
 						if err == nil {
-							err = procToKill.Kill()
+							err = procToKill.SendSignal(syscall.SIGTERM)
```
Samyak2, line 277: I think this should be `SIGKILL` as that is what [`Kill()`](https://pkg.go.dev/github.com/shirou/gopsutil@v3.21.4+incompatible/process#Process.Kill) sends, to maintain consistency

anihm136: I briefly mentioned in the commit message why I chose SIGTERM over SIGKILL, but it is an irrelevant decision either way as it is only temporary. This will eventually be replaced by the list of signals to choose from. We could revisit this if we choose to have a default signal to send.

anihm136:
Do we currently have a way to display a panel/window on one side of the main panel? I'm thinking of a UI similar to htop for this. We could always go for a new overlay window (like the help window), but I feel it would be slightly obstructive to block the process list when the user selects a signal to send. @MadhavJivrajani @Gituser143 let me know your thoughts on this

Gituser143:
We currently have one being used for errors, in `display/misc`. It can be extended to take input and as for the obstruction, we can just render it according to relative terminal dimensions. @anihm136 do you think that could work?

anihm136:
Yep, that seems like what I was looking for!

anihm136:
Okay @Gituser143 I think I did not understand your point about relative terminal dimensions correctly. It appears that the error page is also rendered fullscreen, like the help window. What I'm thinking of is something like this -
image
Is this possible with the currently available stuff?

Gituser143:
@anihm136 Yes. So what you can do is, each drawable widget has a `setRect()` method which can be used to set relative terminal dimensions. For example in your case, you could do:

```go
// Get terminal dimensions 
w, h := ui.TerminalDimensions()

// Set process grid coordinates
page.Grid.SetRect(2*w/3, 0, w, h)

// Set signal table coordinates
page.SigTable.SetRect(0, 0, w/3, h)
```

anihm136:
Got it, thanks!

anihm136:
Okay, the basic signalList is now working. You can open a list of signals, navigate through it and select a signal to send to the process. Some more details are in the commit messages, and I'll start a discussion to decide on a few other aspects of behavior and keybindings

Gituser143:
@anihm136 Instead of inheriting a List widget for the signal list, I feel using the custom table (`/src/utils/table.go`) would be a better choice. This is because it would fit well with the already existing process table and for the reason that it resizes better than a list (If we were to implement more columns in the table, say with the integer codes, list would get messy). What do you think?

MadhavJivrajani:
@anihm136 checking in, how's it going?

anihm136:
Sorry, been a little busy for the last few days. Should be able to make some progress today

anihm136:
@Gituser143 I'll look into using tables instead of a list today

anihm136:
Okay, we now have a table to display the signals (which actually looks a lot nicer and fits the rest of the UI well). I've also added numeric navigation, so you can type in the numeric ID of a signal in the table (as displayed beside it) to select the signal.
I've also taken the liberty of adding a couple of utility functions that I needed for this, as I felt it would be useful elsewhere

src/display/process/allProcs.go
```diff
@@ -268,12 +279,44 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 							killSelected = false
 							myPage.ProcTable.CursorColor = selectedStyle
 						}
+						sendSignal = false
+						updateUI()
 					case "K", "<F9>":
 						// get process and kill it
```
Gituser143, line 285: I'm a little confused, why retain this when we have the `<Enter>` and the numeric cases handled?

src/display/process/allProcs.go
```diff
@@ -268,12 +279,44 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
...
+						ui.Render(signals)
+					case "k", "<Up>":
+						signals.Table.ScrollUp()
+						ui.Render(signals)
+					case "1", "2", "3", "4", "5", "6", "7", "8", "9":
+						scrollIdx, _ := strconv.Atoi(e.ID)
+						if utils.Contains([]string{"1", "2", "3"}, previousKey) != -1 {
+							prevIdx, _ := strconv.Atoi(previousKey)
+							scrollIdx = 10*prevIdx + scrollIdx
+							previousKey = "g" // TODO: don't use this hack
```
Gituser143, line 310: Maybe just set it to an empty string here?

src/display/process/allProcs.go
```diff
@@ -268,12 +279,44 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
...
+						}
+						signals.Table.ScrollToIndex(scrollIdx - 1) // account for 0-indexing
+						ui.Render(signals)
```
Gituser143, line 313: Maybe add a few comments explaining that as number keys are typed, cursor scrolls to that index?

src/utils/slice.go
```diff
@@ -0,0 +1,25 @@
...
+package utils
+
+func Contains(slc []string, item string) int {
+	for i, v := range slc {
+		if v == item {
+			return i
+		}
+	}
+	return -1
+}
```
Gituser143, line 25: This entire thing could be replaced with a map right? it could go like this:
```go
set := map[string]bool{
  "1",  true,
  "2",  true,
}

if _, ok := set[ele]; ok {
...
}
```

It would be better than repeated loops I feel (although it is quite negligible here).

Gituser143 (changes requested):

src/display/process/allProcs.go
Gituser143: Additionally, when this is done, the Kill Table is retained, it doesn't close.

src/display/process/allProcs.go
```diff
@@ -130,6 +133,11 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 		w, h := ui.TerminalDimensions()
 		myPage.Grid.SetRect(0, 0, w, h)
 		help.Resize(w, h)
+		if sendSignal {
+			signals.SetRect(0, 0, w/6, h)
+			myPage.Grid.SetRect(w/6, 0, w, h)
+			ui.Render(signals)
+		}
```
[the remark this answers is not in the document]
anihm136: I made it explicit because the rest of the UI also needs to be rendered after this. I could make it an else-if branch, but would need to repeat the call to render the grid. Or I could have a nested if under the existing else branch to check if the signal menu is open. I felt this was a better alternative. Let me know what you feel is best

src/display/process/allProcs.go
```diff
@@ -223,6 +231,9 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
 					case "G", "<End>":
 						myPage.ProcTable.ScrollBottom()
 					case "K", "<F9>":
+						sendSignal = true
+						updateUI()
+						// println(sendSignal)
```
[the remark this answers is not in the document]
anihm136: Okay, will do. The comment was missed, I'll remove it

src/display/process/allProcs.go
anihm136: As I mentioned in the discussion about keybindings, this was retained as a convenience, since most of the time you want to kill the process. So double tapping K or F9 will do that without having to navigate the menu. However the menu should close after this, will fix that

src/display/process/allProcs.go
anihm136: I will. Need to work on documenting few of the other places as well

src/display/process/allProcs.go
anihm136: That doesn't work, because there is a check at the end of the switch statement that sets it back to e.ID. So we need to refactor to set previousKey only when we know we will check that key later, inside the relevant cases instead of doing it together at the end

src/utils/slice.go
anihm136: Do you mean in allProcs, where I use this? In that case yes, I see now that I can replace it with a map like you mention. We could probably keep this function anyway, because I've found it to be generally useful in a few of my other golang projects (unless I've been doing all of them wrong :wink:)

If you meant using the map in the implementation of this function itself, then I'm not sure how you meant to use it

src/display/process/allProcs.go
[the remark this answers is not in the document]
Gituser143: The issue is, we'd want the help page to take precedence, even if pressed when kill is selected. As of now that hinders the UI as both of them get rendered.

Besting under the existing else branch is a good idea though. Go for it!

src/display/process/allProcs.go
Gituser143: I'm a little confused, but as long as you've got an approach in mind that you think works, should be fine, do try it out.

src/utils/slice.go
Gituser143: Yep, the former is what I meant, the function seemed unnecessary. As for retaining it, it seems fine, but I'd like to not keep an unused function :P

We'll worry about whether to retain it or not later. Nevermind.

CC @MadhavJivrajani

src/display/process/allProcs.go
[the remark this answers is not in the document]
anihm136: Okay this is fixed now. I also had to change when updateUI is called for the help window keybinding, because for some reason handling the same event twice wasn't working well with the signal selector (exiting vie <Esc> was fine, but exiting via "?" was not)

src/display/process/allProcs.go
anihm136: Can you check if the changes I've made to this now make sense?

MadhavJivrajani (commented):
Everything looks great to me! Thanks for your work @anihm136!

I had one suggestion/question: do you think we should reword the `help` menu in `grofer proc`, it says "Kill Process", maybe we can reword it to "Send signal to process"? Wdyt?

Gituser143:
@anihm136 @MadhavJivrajani do you think we should add sorting functionality to the signal table? Considering it already has the option to be navigated purely with the the numbers, I'm not entirely sure. What do y'all think?

anihm136:
@MadhavJivrajani yes it does make sense to reword the help message. I'll do it as part of documentation update

anihm136:
@Gituser143 I don't think sorting makes sense in this context, because generally you aren't looking for signals in any particular order. Two things that would help are -
1. Indexing by signal number. The challenge with this is that different platforms have slightly different numbering for each signal, so I wanted to consider that before implementing
2. Searching. I think this would be a great feature in general, and here in particular it would be helpful because you know either the name or the number of the signal. Between these two you would be able to quickly get to the signal you want

Gituser143:
Isn't it already sorted on signal integers though? We don't have to explicitly do it right?

anihm136:
> Isn't it already sorted on signal integers though? We don't have to explicitly do it right?

They aren't sorted on signal integers now. Indexing is arbitrary (based on order of definition in the syscall package)

src/display/process/allProcs.go
```diff
@@ -283,11 +295,46 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
...
 						updateProcs()
+						sendSignal = false
+						updateUI()
+					case "j", "<Down>":
+						signals.Table.ScrollDown()
+						ui.Render(signals)
+					case "k", "<Up>":
+						signals.Table.ScrollUp()
+						ui.Render(signals)
+					case "1", "2", "3", "4", "5", "6", "7", "8", "9":
```
Samyak2, line 306: I noticed an issue with the numeric navigation. `11`, `19`, etc. all work but `10` does not. It stays at 1. Similarly for `20` and `30`.

Adding `"0"` to this case fixes that

Samyak2 (changes requested):

Samyak2:
I noticed another issue that can be reproduced with the following steps:
1. Run `grofer proc`
2. Run a different process (for example, `cat`) in another terminal such that *it is the last process in the list*
3. Kill it with `KK`
4. Notice no process is currently highlighted
5. Press `K` again to open the signal list
6. Try to navigate the signal list with numbers or `jk`, the navigation happens in the process list instead. `<Escape>` also does not close the signal list. The signal list is stuck there.

I'm not sure what's causing this yet.

src/display/process/allProcs.go
anihm136: Done!

anihm136:
@Samyak2 I am able to reproduce the second issue. Two things to note - 
1. The disappearing highlight seems to be a more general issue. When the last process in the list is highlighted and it dies (for any reason), no process is highlighted. Using the nav keys brings back the highlight to a process in the table
2. After step 5, navigating the process list and pressing `K` on a selected process behaves as expected (does not open a new menu, navigation shifts to the existing menu and selecting a signal sends it to the process and closes the menu)

I think the first issue should be worked on as a separate issue. Meanwhile, I can fix it temporarily by making sure a process is highlighted before opening the menu

anihm136:
Can you verify if the issue is solved now?

Samyak2:
Yes, the issues are resolved now :tada: 

Opening a separate issue for last process highlight sounds good. I think I first noticed it in #54 where it gave an out of range panic.
Would you be able to open the issue for this?

anihm136:
Will do

anihm136:
Okay I think I'm done. Can I have another review to make sure everything's in order?

src/display/misc/keybinds.go
```diff
@@ -40,8 +40,15 @@ var procKeybindings = []string{
 	"  - 0: Disable Sort",
 	"",
 	"[Process actions](fg:white)",
-	"[Requires confirmation, press key again to confirm](fg:white)",
-	"  - K and <F9>: Kill process",
+	"  - K and <F9>: Open signal selector menu",
+	"",
+	"[Signal selection](fg:white)",
+	"  - K and <F9>: Send SIGTERM to selected process",
```
Samyak2, line 46: (nitpick) maybe mention that this kills the process. People not familiar with unix signals may not understand this.

src/display/process/allProcs.go
```diff
@@ -283,11 +296,52 @@ func AllProcVisuals(dataChannel chan []*proc.Process,
...
+						/*
+						* The signal selector can be navigated by entering the number beside the
+						* desired signal. Double digit numbers are handled by checking the previous
+						* key and, if it is among 1,2 and 3, navigate to the corresponding double
+						* digit number (as there are currently 31 supported signals).
+						* For example, pressing 25 would first navigate to signal 2, then to signal 25
+						 */
+						scrollIdx, _ := strconv.Atoi(e.ID)
+						if _, checkPrev := map[string]bool{"1": true, "2": true, "3": true}[previousKey]; checkPrev {
+							prevIdx, _ := strconv.Atoi(previousKey)
```
Samyak2, line 317: (nitpick) since it's just comparing with 3 numbers, won't a simple if-statement be better here?

Converting it to an int before the comparison will also be (negligibly) faster.

```suggestion
						prevIdx, _ := strconv.Atoi(previousKey)
						if prevIdx == 1 || prevIdx == 2 || prevIdx == 3 {
							
```

IMO it also looks cleaner.

Samyak2 (approved):
LGTM. This is great!

Just a couple of small nitpicks.

src/display/misc/keybinds.go
anihm136: Makes sense. Will do

src/display/process/allProcs.go
anihm136: Yeah it does. This was a messy solution to my attempt to do this like I would in Python :smile:. One thing to consider, though, is that we may add more signals in the future. `htop`, for example, has 62 signals (I think?). Not sure if this approach would look as clean for 3 more cases (or the current approach, for that matter)

Samyak2: > `htop`, for example, has 62 signals

Woah. TIL. I never realized the signal window was scrollable xD. Looks like it has 64 signals.

You're right, this won't look very good when there's 6 of them.

Samyak2:
@Gituser143 could you review the PR again? (or @MadhavJivrajani)

MadhavJivrajani (commented):
Left a comment, but looks gooooooooooood to me, thanks again @anihm136 for working on this ❤️ 
I love how it turned out :)

MadhavJivrajani (approved):

MadhavJivrajani:
WOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOT

PR #116, opened 2021-05-16
Remove `mpstat` as a dependency for the CPU Info page

### LINKED ISSUE #97: -c flag throws error
**Describe the bug**
mpstat binary not installed by default.
-c flag throws error "Error: exec: "mpstat": executable file not found in $PATH"

**To Reproduce**
1. Open terminal (shell used: zsh 5.8)
2. Command: grofer -c
3. See error: "Error: exec: "mpstat": executable file not found in $PATH"

**Expected behavior**
Show Info about the CPU Load over all CPUs

**Desktop:**
 - OS: Ubuntu 20.10 x86_64
 - Go Version: 1.14.7

panzerox123:
# Description

Fixes #97 

1. Read `/proc/stat` to obtain CPU load information instead of relying on `mpstat`. No more dependency related errors on systems that don't have `mpstat` installed by default.

2. Updated CONTRIBUTING.MD to say `main` instead of `master` branch.

Let me know if there are any changes you would like :D 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [X] I have read the [contribution guidelines](https://github.com/pesos/grofer/blob/master/CONTRIBUTING.md) and followed it as far as possible. 
- [X] I have performed a self-review of my own code (if applicable)
- [X] I have commented my code, particularly in hard-to-understand areas (if applicable)
- [X] I have run `go fmt` on my code ([reference](https://blog.golang.org/gofmt))
- [X] My changes generate no new warnings 
- [ ] Any dependent and pending changes have been merged and published

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 - Look for issues in the repository. Identify an issue you would like to work on.
 - Once you have done that, leave a comment on the issue saying that you want to work on it. The maintainers will give you the go-ahead.
-- Fork the repository. Your changes will be made on its `master` branch, or on the appropriate feature branch if applicable
+- Fork the repository. Your changes will be made on its `main` branch, or on the appropriate feature branch if applicable
 - Clone the fork on your system. Add a remote to the original `pesos/grofer` repository called `upstream`
 ```
 git clone https://github.com/<your-github-username>/grofer.git
@@ -18,7 +18,7 @@ git commit
 
 - Before pushing your changes, make sure that the changes from upstream are included (use `--rebase` to make sure that your changes stay on top of the latest changes in the upstream repository)
 ```
-git pull --rebase upstream main     # if you are working on a feature branch, use that branch name instead of master
+git pull --rebase upstream main     # if you are working on a feature branch, use that branch name instead of main
 ```
 
 - In the *extremely* small chance that you run into a conflict, just open the files having the conflict and remove the markers and edit the file to the one you want to push. After editing, run `git rebase --continue` and repeat till no conflict remains
@@ -26,16 +26,16 @@ git pull --rebase upstream main     # if you are working on a feature branch, us
 - Verify that your program builds and passes all the tests, and your change actually works in general
 - Push your changes to your fork
 ```
-git push origin main      # if you are working on a feature branch, use that branch name instead of master
+git push origin main      # if you are working on a feature branch, use that branch name instead of main
 ```
-- Visit your forked repository and click on "Pull Request". The Pull Request must always be made to the `pesos/master` branch. Add the relevant description. At this point your name will be assigned to the original issue.
+- Visit your forked repository and click on "Pull Request". The Pull Request must always be made to the `pesos/main` branch. Add the relevant description. At this point your name will be assigned to the original issue.
 - The maintainers will review your code and see if it is okay to merge. It is quite normal for them to suggest you to make some changes in this review.
 - if you are asked to make changes, all you need to do is:
 ```
 # make your change
 git add <files that you changed>
 git commit
-git push origin main      # if you are working on a feature branch, use that branch name instead of master
+git push origin main      # if you are working on a feature branch, use that branch name instead of main
 ```
 - The changes are immediately reflected in the pull request. Once the maintainers are satisfied, they will merge your contribution :)
 
@@ -45,6 +45,6 @@ As long as you follow the above instructions things should go well. You are alwa
 
 - main branch for development. Small patches/enhancements go here.
 - stable branch for tagged releases. This is the branch that will be shipped to users.
-- Separate feature-x branches for adding new "big" features. These branches are merged with master, on completion.
-- Once we are satisfied with a certain set of features and stability, we pull the changes from master to stable. A new release tag is made.
-- If bugs were found on the stable release, we create a hotfix branch and fix the bug. The master branch must also pull the changes from hotfix. A new release tag is created (incrementing with a smaller number).
+- Separate feature-x branches for adding new "big" features. These branches are merged with main, on completion.
+- Once we are satisfied with a certain set of features and stability, we pull the changes from main to stable. A new release tag is made.
+- If bugs were found on the stable release, we create a hotfix branch and fix the bug. The main branch must also pull the changes from hotfix. A new release tag is created (incrementing with a smaller number).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ git commit
 
 - Before pushing your changes, make sure that the changes from upstream are included (use `--rebase` to make sure that your changes stay on top of the latest changes in the upstream repository)
 ```
-git pull --rebase upstream master     # if you are working on a feature branch, use that branch name instead of master
+git pull --rebase upstream main     # if you are working on a feature branch, use that branch name instead of master
 ```
 
 - In the *extremely* small chance that you run into a conflict, just open the files having the conflict and remove the markers and edit the file to the one you want to push. After editing, run `git rebase --continue` and repeat till no conflict remains
@@ -26,7 +26,7 @@ git pull --rebase upstream master     # if you are working on a feature branch, 
 - Verify that your program builds and passes all the tests, and your change actually works in general
 - Push your changes to your fork
 ```
-git push origin master      # if you are working on a feature branch, use that branch name instead of master
+git push origin main      # if you are working on a feature branch, use that branch name instead of master
 ```
 - Visit your forked repository and click on "Pull Request". The Pull Request must always be made to the `pesos/master` branch. Add the relevant description. At this point your name will be assigned to the original issue.
 - The maintainers will review your code and see if it is okay to merge. It is quite normal for them to suggest you to make some changes in this review.
@@ -35,7 +35,7 @@ git push origin master      # if you are working on a feature branch, use that b
 # make your change
 git add <files that you changed>
 git commit
-git push origin master      # if you are working on a feature branch, use that branch name instead of master
+git push origin main      # if you are working on a feature branch, use that branch name instead of master
 ```
 - The changes are immediately reflected in the pull request. Once the maintainers are satisfied, they will merge your contribution :)
 
@@ -43,7 +43,7 @@ As long as you follow the above instructions things should go well. You are alwa
 
 ## Release overview (for the more regular contributors)
 
-- master branch for development. Small patches/enhancements go here.
+- main branch for development. Small patches/enhancements go here.
 - stable branch for tagged releases. This is the branch that will be shipped to users.
 - Separate feature-x branches for adding new "big" features. These branches are merged with master, on completion.
 - Once we are satisfied with a certain set of features and stability, we pull the changes from master to stable. A new release tag is made.
--- a/src/general/cpuInfo.go
+++ b/src/general/cpuInfo.go
@@ -58,18 +58,16 @@ func readCPULoad(c *CPULoad) error {
 	// First line should store in the format:
 	// "cpu <usr> <nice> <system> <idle> <iowait> <irq> <softirq> <steal> <guest> <guest_nice>"
 	vals := strings.Fields(lines[0])
+	// Start from index 1 as first element is the cpu/cpu<no>
+	vals = vals[1:]
 	var avg [10]float64
 	sum := 0
 	for i, x := range vals {
-		// Start from index 1 as first element is the cpu/cpu<no>
-		if i < 1 {
-			continue
-		}
 		curr, err := strconv.Atoi(x)
 		if err != nil {
 			return error
 		} else {
-			avg[i-1] = float64(curr)
+			avg[i] = float64(curr)
 			sum += curr
 		}
 	}
--- a/src/general/cpuInfo.go
+++ b/src/general/cpuInfo.go
@@ -48,23 +48,26 @@ func NewCPULoad() *CPULoad {
 }
 
 // ReadCPULoad reads /proc/stat and returns the average load
-func ReadCPULoad() ([10]float64, error) {
+func readCPULoad(c *CPULoad) error {
 	data, error := os.ReadFile("/proc/stat")
 	if error != nil {
-		return [10]float64{}, error
+		return error
 	}
-	string_data := string(data)
-	lines := strings.Split(string_data, "\n")
+	stringData := string(data)
+	lines := strings.Split(stringData, "\n")
+	// First line should store in the format:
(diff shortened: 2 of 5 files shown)


Gituser143:
@panzerox123 Well that was fast xD

src/general/cpuInfo.go
```diff
@@ -47,29 +47,51 @@ func NewCPULoad() *CPULoad {
 	return &CPULoad{}
 }
 
+// ReadCPULoad reads /proc/stat and returns the average load
+func ReadCPULoad() ([10]float64, error) {
+	data, error := os.ReadFile("/proc/stat")
```
Gituser143, line 52: Could this be made into a slice instead of an array? If not, maybe as a struct as it would make the fields more readable.

src/general/cpuInfo.go
```diff
@@ -47,29 +47,51 @@ func NewCPULoad() *CPULoad {
...
+	if error != nil {
+		return [10]float64{}, error
+	}
+	string_data := string(data)
```
MadhavJivrajani, line 56: `string_data` could be made `stringData` -> not wrong as such, but more like a Go good practice

src/general/cpuInfo.go
```diff
@@ -47,29 +47,51 @@ func NewCPULoad() *CPULoad {
...
+	lines := strings.Split(string_data, "\n")
+	vals := strings.Split(lines[0], " ")
+	var avg [10]float64
+	sum := 0
+	for i, x := range vals {
+		if i < 2 {
```
MadhavJivrajani, line 62: Why `i < 2`?
A comment explaining that might help?

panzerox123:
> @panzerox123 Well that was fast xD

Yeah lol I only asked to be assigned once I figured it out xD

src/general/cpuInfo.go
```diff
@@ -47,29 +47,51 @@ func NewCPULoad() *CPULoad {
...
+			continue
+		}
```
Gituser143, line 64: Why start from index 2? Isn't the line of the format `cpu <user> <nice> <system> ... `?

CONTRIBUTING.md
```diff
@@ -2,7 +2,7 @@
 
 - Look for issues in the repository. Identify an issue you would like to work on.
 - Once you have done that, leave a comment on the issue saying that you want to work on it. The maintainers will give you the go-ahead.
-- Fork the repository. Your changes will be made on its `master` branch, or on the appropriate feature branch if applicable
+- Fork the repository. Your changes will be made on its `main` branch, or on the appropriate feature branch if applicable
 - Clone the fork on your system. Add a remote to the original `pesos/grofer` repository called `upstream`
```
Gituser143, line 6: Completely missed these. Thanks for catching it!

Gituser143 (changes requested):

MadhavJivrajani (commented):
I like the overall approach! It also removes dependency on `gjson`!
And thanks for updating CONTRIBUTING.md! ❤️ 

I have a few suggestions:
- I see that you're returning a 10 element array of floats and then explicitly assigning values 
  What might be cleaner and more readable is having a seperate type, where the component elements are `int` (you could convert them to `int`s) in `ReadCPULoad` iteself so that the code looks as readable and intuitive as possible. 
Something like:
```go
type cpuLoad struct {
    usr int
    nice int
    ....
}
```
And then in `UpdateCPULoad` you can do:
```go
c.Usr = stats.usr
c.Nice = stats.nice
...
```

src/general/cpuInfo.go
panzerox123: Sure I'll update it tomorrow. Never really used Go before so I need to figure out what a slice is xD

src/general/cpuInfo.go
panzerox123: Ah yes. It kept crashing, then I realised there's 2 spaces and not one between CPU and <user>

panzerox123:
Thanks for the reviews guys! I'll make the necessary changes and push another commit tomorrow!

src/general/cpuInfo.go
panzerox123: Got it!

src/general/cpuInfo.go
panzerox123: Sure. It's simply because there's an extra space between CPU and <usr> on the first line of the file. I'll add a comment

src/general/cpuInfo.go
```diff
@@ -47,29 +47,51 @@ func NewCPULoad() *CPULoad {
 	return &CPULoad{}
 }
 
+// ReadCPULoad reads /proc/stat and returns the average load
+func ReadCPULoad() ([10]float64, error) {
```
MadhavJivrajani, line 51: Also, maybe `ReadCPULoad` -> `readCPULoad` to make it non-exportable, I'm assuming this is a helper function not being used outside this package.

panzerox123:
Would it be okay if I just passed a pointer to `c` from `UpdateCPULoad`, instead of creating a new type which will mostly consist of the same data?

src/general/cpuInfo.go
panzerox123: Sure. Will do.

src/general/cpuInfo.go
```diff
@@ -48,23 +48,26 @@ func NewCPULoad() *CPULoad {
...
+		return error
 	}
-	string_data := string(data)
-	lines := strings.Split(string_data, "\n")
+	stringData := string(data)
+	lines := strings.Split(stringData, "\n")
+	// First line should store in the format:
+	// "cpu <usr> <nice> <system> <idle> <iowait> <irq> <softirq> <steal> <guest> <guest_nice>"
 	vals := strings.Split(lines[0], " ")
 	var avg [10]float64
```
Gituser143: Maybe use fields instead? Doc [here](https://golang.org/pkg/strings/#Fields).

Edit: This should make iterating over the values simple too as you can slice and start from first index.

Gituser143 (changes requested):

src/general/cpuInfo.go
```diff
@@ -47,30 +47,56 @@ func NewCPULoad() *CPULoad {
...
+	data, error := os.ReadFile("/proc/stat")
+	if error != nil {
+		return error
+	}
+	stringData := string(data)
+	lines := strings.Split(stringData, "\n")
+	// First line should store in the format:
+	// "cpu <usr> <nice> <system> <idle> <iowait> <irq> <softirq> <steal> <guest> <guest_nice>"
+	vals := strings.Split(lines[0], " ")
+	var avg [10]float64
```
Gituser143, line 61: Maybe use fields instead? Doc [here](https://golang.org/pkg/strings/#Fields).

Edit: This should make iterating over the values simple too as you can slice and start from first index.

Gituser143 (changes requested):

Gituser143:
> Would it be okay if I just passed a pointer to `c` from `UpdateCPULoad`, instead of creating a new type which will mostly consist of the same data?

Yep this is good!

src/general/cpuInfo.go
panzerox123: Done!

src/general/cpuInfo.go
```diff
@@ -47,30 +47,54 @@ func NewCPULoad() *CPULoad {
 	return &CPULoad{}
 }
 
+// ReadCPULoad reads /proc/stat and returns the average load
+func readCPULoad(c *CPULoad) error {
+	data, error := os.ReadFile("/proc/stat")
```
MadhavJivrajani, line 52: `error` is a type name as well. Maybe rename this to something like `err`?

Gituser143 (changes requested):
Since readCPULoad() takes only the pointer to CPULoad as an argument, it might just be better to make it a receiver function.

That can be done by simply declaring it as:
```go
func (c *CPULoad) readCPULoad() err {
}
```

MadhavJivrajani:
A few more suggestions coming your way @panzerox123!

Gituser143:
@panzerox123 Before along with your next commit, do run `go mod tidy` once. It should remove the unnecessary dependencies and I'm hoping it fixes the failing checks at the `go get` step :P

panzerox123:
> @panzerox123 Before along with your next commit, do run `go mod tidy` once. It should remove the unnecessary dependencies and I'm hoping it fixes the failing checks at the `go get` step :P

Cool! And should I include the go.mod and go.sum files as well in the next commit?

Gituser143:
> Cool! And should I include the go.mod and go.sum files as well in the next commit?
Yep! They aren't git ignored or anything, so make sure to push them.

MadhavJivrajani (commented):
I think @Gituser143 already covered the reciever function part

So looking at your code got me thinking. 
- You're making use of only the first line of `/proc/stat`
- To get this line, you're first explicitly converting to `string` and then doing a `Split()`
- One thing that I got concerned about was that the size of `/proc/stat` can grow arbitrarily with cores, interrupts supported on that machine and so on. So what that essentially means is that you would be reading all of this content in memory when you essentially only need the first line. 

### I tried two ways:

---
- First is, since you're reading using `os.ReadFile`, as a result, you'd get a slice of `bytes`, so instead of doing a `Split()` on that, iterate through the slice till you encounter the `byte` value of `\n` and then do
```go
vals := strings.Fields(string(data[:firstNewLineIndex]))
```
This is slightly faster since you're avoiding the `Split()`, but you're still reading out the entire file. 

- The next one is to read the file only till the first `\n` character. You can do this using the [`bufio`](https://golang.org/pkg/bufio/) package. This gives a _significant_ improvement in performance.
```go
const (
        // value of newline in bytes
        newLine = byte(10)
)

func readCPULoad() error {
        ...
        file, error := os.Open("/proc/stat")
        defer file.Close()
        if error != nil {
                return error
        }
        reader := bufio.NewReader(file)

        data, err := reader.ReadBytes(newLine)
        vals := strings.Fields(string(data))
        ...
}
```

Here are the benchmarks for the different approaches:

- `BenchmarkReadCPULoadSplit-8` - The current implementation
- `BenchmarkReadCPULoadBytes-8` - The first approach I mentioned
- `BenchmarkReadCPULoadReader-8` - The second approach using `bufio` 
```
goos: linux
goarch: amd64
pkg: github.com/MadhavJivrajani/test
cpu: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
BenchmarkReadCPULoadReader-8      100000             16606 ns/op            4520 B/op          7 allocs/op
BenchmarkReadCPULoadBytes-8       100000             26001 ns/op           20024 B/op         14 allocs/op
BenchmarkReadCPULoadSplit-8       100000             27805 ns/op           25112 B/op         15 allocs/op
PASS
ok      github.com/MadhavJivrajani/test 7.045s
```

The results are averaged over `100000` runs of each of these approaches. 

- `ns/op` is how long it took on average per approach. 
- `B/op` is how much memory was consumed per function call
- `allocs/op` is the number of _distinct_ memory locations that were used during each approach. Something that the garbage collector would have to handle.

MadhavJivrajani:
Also if you want to play around with the benchmark:
- [`main.go`](https://play.golang.org/p/0BW5vq7eoD3)
- [`main_test.go`](https://play.golang.org/p/n0EXi4UBscz)

```
# run the benchmark
go test -bench=. -benchtime=100000x -benchmem
```

src/general/cpuInfo.go
panzerox123: That's the return type, so it would be a typename right? UpdateCPULoad and GetCPULoad also use `error`

panzerox123:
Yeah let me try these out!

src/general/cpuInfo.go
MadhavJivrajani: Oh no that's not what I meant, I meant changing `data, error := ...` to `data, err := ...`

panzerox123: Ahhh nevermind lol

panzerox123:
Looks like the `go get` error seems to be fixed as well

src/general/cpuInfo.go
```diff
@@ -47,30 +48,53 @@ func NewCPULoad() *CPULoad {
 	return &CPULoad{}
 }
 
+// ReadCPULoad reads /proc/stat and returns the average load
+func (c *CPULoad) readCPULoad() error {
+	file, err := os.Open("/proc/stat")
+	if err != nil {
+		return err
+	}
+	defer file.Close()
```
MadhavJivrajani, line 57:
```suggestion
	file, err := os.Open("/proc/stat")
	defer file.Close()
	if err != nil {
		return err
	}
```

MadhavJivrajani (changes requested):
- This one's probably on me, the files I linked had that as well, sorry about that.
- Can you add comments in the relevant places? Ex: why `[1:]` and what each for loop is supposed to be doing, etc. 

LGTM otherwise, thanks for your work!

src/general/cpuInfo.go
panzerox123: Yeah, it did warn me that I should probably check for the error before calling the file.Close() function, which I why I have the if conditions first. 

Yep. I shall add comments tomorrow!

MadhavJivrajani: Warned you for having `defer` before the `if`?

panzerox123: No the warning only highlighted 'file.close()'. It wasn't a compiler warning, it was the linter

Gituser143: @MadhavJivrajani you only want the defer to be called when `err != nil`, else when error is not nil and we return the error, the close function will be called. This could cause a panic as you're trying to close a file which errored when opening. So I'd avoid changing this. It's good as is.

MadhavJivrajani: Ah I see, now that I think about it, the way you've done it is fine, if there is an error, file wouldn't be open anyway and the Go blog also uses it in a similar way you have

MadhavJivrajani: > 

@MadhavJivrajani you only want the defer to be called when err != nil, else when error is not nil and we return the error, the close function will be called. This could cause a panic as you're trying to close a file which errored when opening. So I'd avoid changing this. It's good as is.

Makes sense.

Gituser143 (approved):

MadhavJivrajani:
Will merge after addition of comments.

Gituser143:
@panzerox123 @MadhavJivrajani sorry for the weird branch updates. Since CI builds weren't triggering here for some reason, I tried setting them up through actions. :P

MadhavJivrajani (approved):
tenor

MadhavJivrajani:
@Gituser143 build is stuck again 🤣

MadhavJivrajani:
I'm gonna merge this considering the previous commit passed all tests and the build was successful and this commit is just addition of comments.

PR #119, opened 2021-05-24
Changed trillion to billion and edge case is taken care

### LINKED ISSUE #118: [BUG] Incorrect unit in `RoundValues()`
**Describe the bug**
In the `utils` package, the `RoundValues()` function has a small error where the switch case classifies units as `per thousand`, `per million` and **`per trillion`**. Additionally, no default case is set, which could lead to the function returning empty values.

**To Reproduce**
Unsure how to reproduce.

**Expected behaviour**
The third unit must be `per billion` and not `per trillion`

 - OS: Ubuntu 18.04
 - Go Version 1.16

mihirkestur:
For the edge case (n >= T), numbers will be represented in terms of "gB" units. i.e. All numbers greater than or equal to 10^9 will be represented in terms of gigabytes (billion bytes). This resolves the edge case, if numbers were greater than 10^12, empty values would be returned.

Fixes #118 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [X] I have read the [contribution guidelines](https://github.com/pesos/grofer/blob/master/CONTRIBUTING.md) and followed it as far as possible. 
- [ ] I have performed a self-review of my own code (if applicable)
- [ ] I have commented my code, particularly in hard-to-understand areas (if applicable)
- [ ] I have run `go fmt` on my code ([reference](https://blog.golang.org/gofmt))
- [X] My changes generate no new warnings 
- [ ] Any dependent and pending changes have been merged and published

### DIFF
--- a/src/utils/dataFormat.go
+++ b/src/utils/dataFormat.go
@@ -25,6 +25,8 @@ var (
 	K = math.Pow(10, 3)
 	M = math.Pow(10, 6)
 	G = math.Pow(10, 9)
+	T = math.Pow(10, 12)
+	Q = math.Pow(10, 15)
 )
 
 func roundOffNearestTen(num float64, divisor float64) float64 {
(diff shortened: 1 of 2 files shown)


src/utils/dataFormat.go
```diff
@@ -25,7 +25,6 @@ var (
 	K = math.Pow(10, 3)
 	M = math.Pow(10, 6)
 	G = math.Pow(10, 9)
-	T = math.Pow(10, 12)
 )
```
Gituser143, line 28: @mihirkestur Although this makes sense, maybe it would be better to retain T (Trillion, Peta Bytes), maybe making Peta Bytes as the last case would be better. I know it's unlikely to have metrics in pB, but it is a possibility.

Gituser143 (changes requested):

src/utils/dataFormat.go
mihirkestur: Alright sure!

Gituser143 (approved):
Looks good to me! I'll let the next reviewer merge this.
CC @MadhavJivrajani @Samyak2

MadhavJivrajani (approved):
Thanks for your contribution 🎉 ❤️

PR #124, opened 2021-06-05
Add Dockerfile pt. 2

### LINKED ISSUE #101: [FEATURE REQ] Add Dockerfile for `grofer`

Samyak2:
# Description

Fixes #101 

Based on #106 

 - Added a Dockerfile based on the [`golang` image](https://hub.docker.com/_/golang)
 - `grofer` is installed using `go install` from the local source
 - The whole root directory of the host has to be mounted as read only for this to work. A few other docker flags are also required.
    - the docker socket is also mounted
 - See updated README for instructions
 - `HOST_PROC`, `HOST_SYS`, etc. is set for `gopsutil`

## Type of change

# Checklist:

# TODO

 - fix GitHub workflow to publish container

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,11 @@ go build grofer.go
    ```
    docker pull ghcr.io/pesos/grofer
    ```
+
+   Note: This will use the `latest` image which is built from the main branch.
+   For stable releases, you will need to use the version number as the tag.
+   Example: `docker pull ghcr.io/pesos/grofer:1.3.0`
+
 1. Run using
    ```bash
    docker run \
--- a/README.md
+++ b/README.md
@@ -52,20 +52,20 @@ go build grofer.go
    ```
 1. Run using
    ```bash
-   docker run --name grofer --rm -it -v /:/host:ro grofer
+   docker run --name grofer --rm -it -v /:/host:ro --privileged --pid=host grofer
    ```
 
    Additional command to `grofer` can be passed directly at the end of the above command:
    ```bash
-   docker run --name grofer --rm -it -v /:/host:ro grofer proc
+   docker run --name grofer --rm -it -v /:/host:ro --privileged --pid=host grofer proc
    ```
 
    ```bash
-   docker run --name grofer --rm -it -v /:/host:ro grofer --help
+   docker run --name grofer --rm -it -v /:/host:ro --privileged --pid=host grofer --help
    ```
 1. (Optional) create an alias for `docker run` to be able to run it using just `grofer`
    ```zsh
-   alias grofer="docker run --name grofer --rm -it -v /:/host:ro grofer"
+   alias grofer="docker run --name grofer --rm -it -v /:/host:ro --privileged --pid=host grofer"
    ```
 
    Then use `grofer` as normal:
--- a/README.md
+++ b/README.md
@@ -52,20 +52,20 @@ go build grofer.go
    ```
 1. Run using
    ```bash
-   docker run --name grofer --rm -it -v /:/rootfs:ro grofer
+   docker run --name grofer --rm -it -v /:/host:ro grofer
    ```
 
    Additional command to `grofer` can be passed directly at the end of the above command:
    ```bash
-   docker run --name grofer --rm -it -v /:/rootfs:ro grofer proc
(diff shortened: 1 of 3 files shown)


Samyak2:
```
unsupported: GitHub Container Registry is currently in read only mode. Image pushes are disallowed at this time.
```

Looks like the container registry is [under maintenance](https://www.githubstatus.com/) currently :(

Samyak2:
It wooooorks!

https://github.com/orgs/pesos/packages/container/package/grofer

MadhavJivrajani:
WOOOHOOOOOOO

README.md
```diff
@@ -44,6 +44,82 @@ cd grofer
 go build grofer.go
 ```
 
+### Docker
+
+1. Get docker image using
+   ```
+   TODO: docker image is not uploaded yet
```
MadhavJivrajani, line 51: Now I guess we can fill this in 🎉

Samyak2: Yes, I'm doing that rn.

I can't exactly test it since `latest` isn't pushed yet. Pulling using `pr-124` tag works though.

Samyak2:
I think this is ready now. I made a few changes
 - Docker workflow is disabled for PRs since it will fail unless it's from an internal branch (like this)
 - Changed the README to use `ghcr.io/pesos/grofer` instead of `grofer`

Gituser143:
@Samyak2 you might want to update the branch :P

Samyak2:
I'll rebase it on main

Samyak2:
Is there anything left to do here?

MadhavJivrajani:
Everything looks good to me, just to be sure: this image build will be triggered only when a new release comes out?

Samyak2:
No. This is how it will work:
- All pushes to main (the default branch) are released as the `latest` image.
- New tags are released with the specific version number. For example, a `1.3.1` tagged image would be pushed on the next release.

So, pulling `ghcr.io/pesos/grofer` would give the image from the main branch (since that is tagged as `latest`). This, if I'm not wrong, is similar to `go get` which pulls from the default branch.

MadhavJivrajani:
Sounds good
Could you add 1-2 lines in the READE along the lines of what `latest` means in our case and if they want to use a particular version, how they can do it?

Samyak2:
I've added a couple of lines on that.

A few things that should be noted here:
 - The `latest` image and release images (tags) are not tested xD. They can only be tested once we merge this (and the tag one when we make a new release). `act` can't test these either since it fails at the login step. We might have to make hotfixes if it breaks. On that note, is it possible to force Actions to run the release (tag) workflow manually so that the `1.3.0` image is built?
 - I'm not entirely sure about the version tag. According to the code (which I ~stole~ borrowed from the [starter workflows](https://github.com/actions/starter-workflows/blob/68d67073c3864b2836a59ed0af5b30d2ad3ff719/ci/docker-publish.yml#L71)) it should be `1.3.0` without the `v` prefix. We might have to change this in the README if it's different.

Samyak2:
^ Rebased on main

Is there anything else to do here?

MadhavJivrajani (approved):
Hey, I missed this
I see there are a few small details regarding version to be worked out, but that can be done independently later, with the documentation in the README, this looks good to me in the current state and can be merged

cc @Gituser143

Gituser143 (approved):

Gituser143:
@Samyak2 
Changes look good to me! But do workflows trigger on commits? I'm asking because I don't see one for this (or I might not be looking properly, plis help :P).

Samyak2:
> @Samyak2 
> Changes look good to me! But do workflows trigger on commits? I'm asking because I don't see one for this (or I might not be looking properly, plis help :P).

@Gituser143 yes, but this one is run only on commits to the main branch and not on PRs. It can be enabled on PRs too, but if they aren't from an internal branch of this repo (pesos/grofer), it will just give a permission denied error.
Also, it's run on push rather than commits. So if multiple commits are pushed at a time, it will run only once for the latest commit.

PR #132, opened 2021-08-20
Clean up github workflows

Gituser143:
# Description
GitHub Workflows have now been split as follows:
- Go Checks (go-checks.yml): Performs checks using go `build`, `fmt`, `vet` and `test`
- License Check (license-check.yml): Performs a license check using `scripts/license_check.py`
- Publish Development Docker Image (docker-publish.yml): Builds and pushes an image upon changes (`push`) to the `main` branch. Image is built and tagged as `pesos/grofer:main` and `pesos/grofer:latest`.
- Publish Stable Docker Image (docker-publish-stable.yml): Builds and pushes only when tags are pushed. After a PR is merged into `stable`, a manual `tag` must be applied to the commit and pushed. Image is then built and tagged as `pesos/grofer:<version>` and `pesos/grofer:latest`.

# Scope of PR:
- [x] docker-publish-stable.yml  
- [x] docker-publish.yml  
- [x] go-checks.yml  
- [x] license-check.yml

# To Do Post PR
- [x] Newer license check in Go? :P 
- [ ] Add Workflow badges to README. [Reference](https://docs.github.com/en/actions/managing-workflow-runs/adding-a-workflow-status-badge)
- [ ] Add Go Report Badge in README. [Reference](https://goreportcard.com/report/github.com/pesos/grofer)
- [x] Add Version Badge in README. Reference [this](https://shields.io/category/version) or [this](https://img.shields.io/badge/<LABEL>-<MESSAGE>-<COLOR>). Latter is customizable.


panzerox123:
LGTM! 

Is there a reason for having different licenses for go and bash files? Both of them seem to be apache2.

Gituser143:
> Is there a reason for having different licenses for go and bash files? Both of them seem to be apache2. 

They're the same license but with the additional comment syntax of the language. For example the go license starts and ends with `/*..*/`, while each line in the bash license starts with `#`.

panzerox123:
Ahh got it!

README.md
```diff
@@ -46,17 +47,17 @@ go build grofer.go
...
 
-   Note: This will use the `latest` image which is built from the main branch.
-   For stable releases, you will need to use the version number as the tag.
-   Example: `docker pull ghcr.io/pesos/grofer:1.3.0`
+1.	Run using
 
-1. Run using
-   ```bash
+```bash
    docker run \
```
Samyak2, line 61: Is there any specific reason for these formatting changes, or was it an auto-formatter?

It breaks the numbering ([see here](https://github.com/pesos/grofer/tree/actions#docker)):
image

Samyak2 (changes requested):
> * Publish Development Docker Image (docker-publish.yml): ... Image is built and tagged as `pesos/grofer:main` and `pesos/grofer:latest`.

According to the [metadata-action README](https://github.com/docker/metadata-action#basic), the latest tag is only used on tag pushes. So, on pushes to `main` a corresponding `main` tagged container is released (and not `latest`). So this should be:
> * Publish Development Docker Image (docker-publish.yml): ... Image is built and tagged as `pesos/grofer:main`.

Is that right?

README.md
Gituser143: Oops, that's an auto formatter. That's on me. Will fix right away.

scripts/license_check.go
```diff
@@ -0,0 +1,80 @@
...
+		})
+
+	if err != nil {
+		log.Println(err)
+	}
+
+	if !verified {
+		os.Exit(1)
+	} else {
+		fmt.Println("All files have license verified.")
```
Samyak2, line 78: nit: you could change this to `log.Println` and remove the `fmt` import.

Otherwise nice work on the script. With this, grofer is officially python-free (I think?) :tada:

Gituser143: Done!

Gituser143:
> > * Publish Development Docker Image (docker-publish.yml): ... Image is built and tagged as `pesos/grofer:main` and `pesos/grofer:latest`.
> 
> According to the [metadata-action README](https://github.com/docker/metadata-action#basic), the latest tag is only used on tag pushes. So, on pushes to `main` a corresponding `main` tagged container is released (and not `latest`). So this should be:
> 
> > * Publish Development Docker Image (docker-publish.yml): ... Image is built and tagged as `pesos/grofer:main`.
> 
> Is that right?

Ah yes, you're right. I misread that table. So now development image will only be tagged as `grofer:main`, while stable ones will have the version tag and `latest`. I'll make the change in the README.

But would we want the image built from `main` to have the `latest` tag? Or would stable versions suffice? If the former, there might a few small changes required.

Samyak2:
> But would we want the image built from `main` to have the `latest` tag?

I wouldn't want that. Generally, `latest` means the latest stable version (don't have a source for this, just what I have observed). A `main` tag for the development branch is good enough.

Gituser143:
> > But would we want the image built from `main` to have the `latest` tag?
> 
> I wouldn't want that. Generally, `latest` means the latest stable version (don't have a source for this, just what I have observed). A `main` tag for the development branch is good enough.

In that case, appropriate changes have been made. Do let me know if I've missed something.

Samyak2 (approved):
LGTM!

PR #134, opened 2021-09-06
Add container actions table and logo!

### LINKED ISSUE #123: [Enhancement] Improve the way container actions are handled
Following #112, the menu of signals to send to processes turned out great and is really user friendly as well, since we have a bunch of actions that we can perform on a container as well through `grofer`, we can probably better the feedback given on execution of an action (currently the color of the row is changed as an indication of confirming that action via executing the same key-binding again).

Since the number of actions is significantly lesser, a menu similar to #112 might not work. We can instead having a floating window come up (similar to help windows that come up when we press `?` on a UI based command), along with details about the container on which the action is to be performed. These details could include but are not limited to:
- ID of the container
- Name of the container
- Image 

Feel free to use this space to discuss and hammer out the specifics of how it should work :)

### LINKED ISSUE #131: [FEATURE REQ] Add an action table for the overall container page
Currently, container actions are performed by pressing specific keys and then pressing them again to confirm the action. Remembering these keys might be quite annoying and it's easy to perform the wrong action. 

A table can be made to pop up much the signal table for the `proc` page which has the list of container actions and the user can select the action to be performed by navigating the table. This eliminates the need to have multiple key bindings for actions and provides a cleaner interface to the user to perform said actions.

Gituser143:
# Description

Added a container actions table. Container actions can now be performed by visually selecting an action instead of remembering the key binding.

Fixes #131 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

# Scope of PR:

- [x] Add logo!
- [x] Add an actions table to `grofer container`
- [x] Lint code (optional)
- [x] Fix Container CPU gauge to give in range of 0-100
- [x] Turn selected utilities into an enum instead of strings 
- [x] Replace error cursor to error box to `proc` page

Edit: Fixes #123 (Duplicate issue)


Samyak2:
I think I found an issue. The help page shows `s` to be bound to pause rendering, but it does not work. The UI and values continue to be updated.

There doesn't seem to be a case for `s` key here:
https://github.com/pesos/grofer/blob/86954fe342c20c58c64535acd6cdd32aab6ee103/pkg/sink/tui/container/overall_container.go#L178

pkg/sink/tui/container/overall_container.go
```diff
@@ -105,6 +107,12 @@ func OverallVisuals(ctx context.Context, cli *client.Client, all bool, dataChann
 			errorBox.Resize(w, h)
 			ui.Render(errorBox)
 
+		case "ACTION":
+			actions.SetRect(0, 0, w/6, h)
+			page.Grid.SetRect(w/6, 0, w, h)
+			ui.Render(actions)
+			ui.Render(page.Grid)
+
```
Samyak2, line 110: Maybe consider making `utilitySelected` an enum? (in the same way it was done for [`Command`](https://github.com/pesos/grofer/blob/20010bd2c7e8eca598adb14d122cbbfcec9593b5/pkg/core/types.go#L19-L33)).

Since it's run on every UI update, it would prevent many string comparisons from taking place every tick.

Is this right, or does the go compiler optimize these things?

Gituser143:
> I think I found an issue. The help page shows `s` to be bound to pause rendering, but it does not work. The UI and values continue to be updated.
> 
> There doesn't seem to be a case for `s` key here:
> 
> https://github.com/pesos/grofer/blob/86954fe342c20c58c64535acd6cdd32aab6ee103/pkg/sink/tui/container/overall_container.go#L178

Oops, I probably missed this in the UI refactor, will fix.

pkg/sink/tui/container/overall_container.go
Gituser143: This does make sense, I'll update for all pages by next commit.

Gituser143: Done.

pkg/sink/tui/misc/keybinds.go
```diff
@@ -145,7 +145,7 @@ func getPerProcCommandKeybindings() [][]string {
 func getContainerCommandKeybindings() [][]string {
 	return [][]string{
 		{"Quit: q or <C-c>"},
-		{"Pause Rendering: s"},
+		{"Pause Rendering: p"},
```
MadhavJivrajani, line 148: Is there a reason for changing `s` to `p` for pause?

README.md
```diff
@@ -1,3 +1,7 @@
+<p align="center">
+  <img src="images/README/logo.png" style="width:300px;" />
+</p>
+
 Grofer
```
MadhavJivrajani, line 5: Can we change title to `grofer`? idk capital G looks weird :P

Samyak2:
I noticed another issue related to pausing. This happens in both `proc` and `containers`. When the UI is paused (using `p`) and you open the help widget (using `?`), when you go back to the main page (proc or containers) (using `Esc`) it gets unpaused automatically.

Is this expected behaviour? I couldn't find anything that would unpause the UI when pressing Esc.

cmd/about.go
```diff
@@ -25,7 +26,7 @@ import (
 )
 
 // groferVersion is the version of grofer that is loaded in during build
-var groferVersion string = "1.3.0"
+var groferVersion = "1.3.0"
```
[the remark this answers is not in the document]
Gituser143: Sounds good

README.md
Gituser143: Fair enough. Will change,

pkg/sink/tui/misc/keybinds.go
Gituser143: Yes, "p" for pause :P

Gituser143:
> I noticed another issue related to pausing. This happens in both `proc` and `containers`. When the UI is paused (using `p`) and you open the help widget (using `?`), when you go back to the main page (proc or containers) (using `Esc`) it gets unpaused automatically.
> 
> Is this expected behaviour? I couldn't find anything that would unpause the UI when pressing Esc.

This is probably because a selection happens to perform an action/kill. `Esc` then deselects it. I'll fix that in the next commit.

Gituser143:
@Samyak2 @MadhavJivrajani changes have been made as per comments, additionally, an error box has been introduced to replace the error cursor colour for the `proc` page. Let me know if I've missed/need to change anything.

Samyak2 (approved):
LGTM

New logo looks amazing :heart:

MadhavJivrajani (approved):
❤️

PR #148, opened 2026-09-19
fix: reset selection cursor when process list shrinks

### LINKED ISSUE #122: [BUG] Selectoed process out of bounds in `grofer proc`
**Describe the bug**
When the process list changes for any reason and is refreshed, if the number of processes in the refreshed list is less than the index of the currently selected process, the selection goes out of bounds and no process is selected

**To Reproduce**
1. `grofer proc`
2. Scroll down to the last process
3. Wait for a while (or kill a process if you're impatient)
4. No process is selected

**Expected behavior**
If the selection goes out of bounds, cursor should reset to the last element in the new list

**Screenshots**
image
Before termination of last process
image
After termination of last process

 - OS: Linux 5.12.7-arch1-1
 - Go Version: 1.16.4 linux/amd64

**Additional context**
Seems generally harmless for now, does not cause any crashes. Using any of the navigation keys from this state resets the selection to a valid position. However, this caused some issues on #112 for which a temporary fix has been made. Updating the selection using `calcPos` after every process refresh seems to fix the problem, but has not been tested for adverse behaviors or side effects

clydeeshun94:
When the process list refreshes with fewer entries (e.g., a process is terminated), SelectedRow could exceed the new row count, leaving no process selected. This fix clamps SelectedRow to the valid range after each refresh.

Fixes #122


MadhavJivrajani (approved):
Amazing, thanks! ♥️

